PR #2, opened 2023-11-14
Initial work for v1 api generation

Yantrio:
Note, `fs.FS` doesn't have write methods on it. So for now we have a hacky read+write set of abstractions.

We should probably look at abstracting it nicely away soon, but I needed this so I can write some tests.

### DIFF
--- a/src/internal/repository-metadata-files/module/list.go
+++ b/src/internal/repository-metadata-files/module/list.go
@@ -28,12 +28,10 @@ func extractModuleDetailsFromPath(path string) *module.Module {
 		return nil
 	}
 
-	// matches[0] is the entire string that matched the regex, so we skip it.
-	// Using named matches in golang is a pain, so we just use the index of the match.
 	m := module.Module{
-		Namespace:    matches[1],
-		Name:         matches[2],
-		TargetSystem: matches[3],
+		Namespace:    matches[moduleDirectoryRegex.SubexpIndex("Namespace")],
+		Name:         matches[moduleDirectoryRegex.SubexpIndex("Name")],
+		TargetSystem: matches[moduleDirectoryRegex.SubexpIndex("TargetSystem")],
 	}
 
 	return &m


src/internal/repository-metadata-files/module/list.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+
+	"registry-stable/internal/module"
+)
+
+func ListModules() ([]module.Module, error) {
+	// TODO: Accept a directory as a parameter (or pass it in some other way) so we can easily test this
```
cube2222, line 12: I recommend just passing as a parameter.

src/internal/repository-metadata-files/module/list.go
```diff
@@ -0,0 +1,41 @@
...
+
+	// walk the module directory recursively and find all json files
+	// for each json file, parse it into a module.Module struct
+	// return a slice of module.Module structs
+
+	var results []module.Module
+	err := filepath.Walk(moduleDirectory, func(path string, info os.FileInfo, err error) error {
+		if filepath.Ext(path) == ".json" {
+			// extract the module details from the file path. It should be modules/<firstLetterOfNamespace>/<namespace>/<name>/<system>.json
+			regex := regexp.MustCompile(`modules\/([a-z])\/([a-z0-9-]+)\/([a-z0-9-]+)\/([a-z0-9-]+).json`)
```
cube2222, line 23: I believe there can be capital letters in namespaces, no? I'm not sure.

cube2222: I believe they were in Arel's PoC, but maybe the github api is case-insensitive.

src/internal/v1api/generator.go
```diff
@@ -0,0 +1,26 @@
...
+// This should take information from the current state of the registry and generate the responses
+// to be served by the API.
+// For more information on the API see the following documentation:
+// - https://opentofu.org/docs/internals/module-registry-protocol/
+// - https://opentofu.org/docs/internals/provider-registry-protocol
+type Generator struct {
+	// DestinationDir is the directory to write the generated responses to.
+	DestinationDir string
+
+	// TODO: Maybe combine all the fs.FS stuff and FileWriter stuff into a single interface to be consumed easier?
```
cube2222, line 19: Honestly, to me this looks like a lot of needless abstraction.

We're operating on the file system, I don't see a reason to mock the file system. Just make all functions accept the directory they should operate on and have the Generator contain the relevant file paths, imo.

src/internal/v1api/modules.go
```diff
@@ -0,0 +1,128 @@
...
+		return fmt.Errorf("failed to write file: %w", err)
+	}
+
+	return nil
+}
+
+// writeModuleVersionDownload writes the file containing the download link for the module version.
+// This data is to be consumed when an end user requests /v1/modules/{namespace}/{name}/{targetSystem}/{version}/download
+func (g Generator) writeModuleVersionDownload(namespace string, name string, system string, version string) interface{} {
+	// the file should just contain a link to GitHub to download the tarball, ie:
```
cube2222, line 110: I am not a fan of this from a perspective of protocol-evolution.

This should imo be a json object with a single field. To stay consistent with the provider protocol, I think it would make sense to use `download_url` here.

In other words, wrap the above into
```
{
  "download_url": "..."
}
```

src/internal/v1api/generator.go
cam72cam: Agreed

src/internal/repository-metadata-files/module/list.go
Yantrio: Good catch, Ill update this regex :)

src/internal/v1api/generator.go
Yantrio: I mainly wanted to mock it so i can test with an in-mem filesystem, and I thought it would be easier to do it from the get-go rather than shoehorning it in later. But I agree, its maybe a bit too much.

src/internal/repository-metadata-files/module/list.go
cube2222: However, maybe we should actually only store lowercase in the repository?

cube2222: I don't think there are uppercase namespaces in the hashi registry?

src/internal/v1api/wellknown.go
```diff
@@ -0,0 +1,31 @@
...
+	err := g.FileWriter.MkdirAll(wellKnownDir, 0755)
+	if err != nil {
+		return fmt.Errorf("failed to create directory: %w", err)
+	}
+
+	filePath := filepath.Join(wellKnownDir, "terraform.json")
+	err = g.FileWriter.WriteFile(filePath, []byte(WellKnownFileContents), 0644)
+	if err != nil {
+		return fmt.Errorf("failed to write file: %w", err)
+	}
```
RLRabinowitz, line 28: **(nit/question)** WDYT about encapsulating the need to `MkdirAll` -> `WriteFile`? As it's done multiple times, and also used when updating the module metadata files

src/internal/v1api/modules.go
RLRabinowitz: +1 on making it a JSON
In my PoC it was `{"X-Terraform-Get": "..."}` but we'll see about the field name later (`download_url` is probably preferable)

src/internal/v1api/modules.go
```diff
@@ -0,0 +1,128 @@
...
+	// list directories at the root of the fs
+	dirs, err := fs.ReadDir(g.ModuleFS, ".")
+	if err != nil {
+		slog.Error("Failed to list directories", slog.Any("err", err))
+		os.Exit(1)
+	}
+
+	for _, d := range dirs {
+		slog.Info("Found directory", slog.String("dir", d.Name()))
+	}
```
RLRabinowitz, line 64: I'm guessing this is only temporary for debug purposes?

src/internal/repository-metadata-files/module/list.go
RLRabinowitz: Modules with capital letters in the namespace should be supported
Also, `_` should be supported

src/internal/repository-metadata-files/module/list.go
```diff
@@ -0,0 +1,41 @@
...
+			matches := regex.FindStringSubmatch(path)
+			if len(matches) == 5 {
+				results = append(results, module.Module{
+					Namespace:    matches[2],
+					Name:         matches[3],
+					TargetSystem: matches[4],
```
RLRabinowitz, line 29: **(nit)** I prefer using named groups here

RLRabinowitz (commented):
Looks good! Left a few comments

src/internal/v1api/modules.go
Yantrio: Oops, yes ill remove this.

src/internal/v1api/modules.go
Yantrio: I've set it to just `location` right now, as i feel that it's generic enough to cover things. Especially as it may not always be a direct download url in the future.

Yantrio: We can change this and discuss once the RFC for moving it to the body has been done

cam72cam (approved):

PR #3, opened 2023-11-14
Add code to create the provider metadata file

RLRabinowitz:
Add code that helps to build and create the provider metadata file

Uses GitHub Releases API to get all the necessary info

Right now - This does not include data necessary from asset downloads (Shasum and protocols)


src/internal/repository-metadata-files/provider/build.go
```diff
@@ -0,0 +1,77 @@
...
+
+	versions := make([]provider.Version, 0)
+	for _, r := range releases {
+		var shaSumsArtifact github.ReleaseAsset
+		var shaSumsSignatureArtifact github.ReleaseAsset
+
+		var targets = make([]provider.Target, 0)
+		for _, asset := range r.ReleaseAssets.Nodes {
+			if platform := github.ExtractPlatformFromFilename(asset.Name); platform != nil {
+				if err != nil {
```
cam72cam, line 32: err is not set anywhere near here

cam72cam (approved):
One minor comment, but it's not causing issues.

Yantrio (approved):
Approved, just @cam72cam's err comment that needs fixing IMHO

PR #7, opened 2023-11-15
Add Protocols and Shasums to the provider metadata file generated

RLRabinowitz:
Download all manifest and shasum files in parallel, you use the information from them to fill the "protocols" and "shasum" parts of the provider metadata file

### DIFF
--- a/src/internal/github/assets.go
+++ b/src/internal/github/assets.go
@@ -15,35 +15,37 @@ type Platform struct {
 	Arch string
 }
 
-func DownloadAssetContents(ctx context.Context, downloadURL string) (body io.ReadCloser, err error) {
+func DownloadAssetContents(ctx context.Context, downloadURL string) ([]byte, error) {
 	httpClient := httpInternal.GetHttpRetryClient()
 
 	log.Printf("Downloading asset, url: %s", downloadURL)
-	req, reqErr := http.NewRequestWithContext(ctx, http.MethodGet, downloadURL, nil)
-	if reqErr != nil {
-		log.Printf("Failed to create request %s", reqErr)
-		err = fmt.Errorf("failed to create request: %w", reqErr)
-		return
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, downloadURL, nil)
+	if err != nil {
+		log.Printf("Failed to create request %s", err)
+		return nil, fmt.Errorf("failed to create request: %w", err)
 	}
 
-	resp, respErr := httpClient.Do(req)
-	if respErr != nil {
-		log.Printf("Error downloading asset %s", respErr)
-		err = fmt.Errorf("error downloading asset: %w", respErr)
-		return
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		log.Printf("Error downloading asset %s", err)
+		return nil, fmt.Errorf("error downloading asset: %w", err)
 	}
 
 	if resp.StatusCode != http.StatusOK {
 		resp.Body.Close()
 		log.Printf("Unexpected status code when downloading asset: %d", resp.StatusCode)
-		err = fmt.Errorf("unexpected status code when downloading asset: %d", resp.StatusCode)
-		return
+		return nil, fmt.Errorf("unexpected status code when downloading asset: %d", resp.StatusCode)
 	}
 
-	body = resp.Body
-
 	log.Printf("Asset downloaded successfully")
-	return
+
+	contents, err := io.ReadAll(resp.Body)
+	resp.Body.Close()
+	if err != nil {
+		return nil, fmt.Errorf("failed to read asset contents: %w", err)
+	}
+
+	return contents, nil
 }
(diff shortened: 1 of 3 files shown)


src/internal/repository-metadata-files/provider/artifacts.go
```diff
@@ -0,0 +1,147 @@
...
+	versionsCopy := versions
+
+	shaSumCh := make(chan ShaSumResult, len(versionsCopy))
+	protocolsCh := make(chan ProtocolsResult, len(versionsCopy))
+
+	var wg sync.WaitGroup
+	for i, v := range versionsCopy {
+		wg.Add(2)
+
+		go func(v provider.Version, i int) {
```
cam72cam, line 99: I think this is fine for now, we probably want to make sure we limit the concurrency as to not hit rate limits as we ramp up.

cam72cam (approved):

src/internal/repository-metadata-files/provider/build.go
```diff
@@ -23,50 +21,45 @@ func BuildMetadataFile(p provider.Provider) (*provider.MetadataFile, error) {
 	}
 
 	versions := make([]provider.Version, 0)
+	versionArtifactsMap := make(VersionArtifactsMap)
+
 	for _, r := range releases {
```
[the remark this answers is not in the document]
RLRabinowitz: I think that makes sense. I'll do that now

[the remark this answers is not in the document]
RLRabinowitz: Done in 73a2f6652df10c00638d0a5847a42a54f0052475

cam72cam (approved):

src/internal/github/assets.go
```diff
@@ -1,14 +1,51 @@
...
 
 type Platform struct {
 	OS   string
 	Arch string
 }
 
+func DownloadAssetContents(ctx context.Context, downloadURL string) (body io.ReadCloser, err error) {
+	httpClient := httpInternal.GetHttpRetryClient()
+
+	log.Printf("Downloading asset, url: %s", downloadURL)
```
Yantrio, line 21: I've raised this for after we merge this PR: https://github.com/opentofu/registry-stable/issues/15

I think we should be using slog everywhere possible

src/internal/github/assets.go
```diff
@@ -1,14 +1,51 @@
...
+	req, reqErr := http.NewRequestWithContext(ctx, http.MethodGet, downloadURL, nil)
+	if reqErr != nil {
+		log.Printf("Failed to create request %s", reqErr)
+		err = fmt.Errorf("failed to create request: %w", reqErr)
```
Yantrio, line 25: Let's clean this up after (//cc @cam72cam / @kislerdm ), but i dont see any reason here to return named properties, 

we could easily just return `nil, err` here.

src/internal/github/assets.go
```diff
@@ -1,14 +1,51 @@
...
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		resp.Body.Close()
+		log.Printf("Unexpected status code when downloading asset: %d", resp.StatusCode)
+		err = fmt.Errorf("unexpected status code when downloading asset: %d", resp.StatusCode)
+		return
+	}
+
+	body = resp.Body
```
Yantrio, line 43: I feel like this code has come across from the old registry,and I was not a fan of it back then because the caller has to be aware that they need to be closing the body. 

Maybe we could now have this return a string/[]byte instead?

src/internal/repository-metadata-files/provider/build.go
```diff
@@ -2,74 +2,125 @@ package provider
...
+	}
+	slices.SortFunc(versions, semverSortFunc)
+
+	return provider.MetadataFile{
+		Repository: oldMetadata.Repository,
+		Versions:   versions,
+	}
+}
+
+func getExistingMetadata(p provider.Provider) (provider.MetadataFile, error) {
```
Yantrio, line 115: For cleaning up after: we should probably have something to read/write metadata that can be shared between this package and the v1 renderer.

cube2222 (approved):

Yantrio (approved):
Mostly nitpicks, but as we're looking to move fast im happy with the logic here and I'll approve assuming we'll clean up after.

src/internal/github/assets.go
RLRabinowitz: Yes I think we'll just have a cleanup task for that

src/internal/github/assets.go
RLRabinowitz: Done in 8abba6412fe3fb6d26eee823a5d2e25f3b55a626

src/internal/github/assets.go
RLRabinowitz: Done in 8abba6412fe3fb6d26eee823a5d2e25f3b55a626

PR #9, opened 2023-11-15
Refactor and add tests for git tag parsing

cam72cam:
This breaks out the git stdout parsing logic into it's own function and add tests around it.  We could add a test for exec-git at some point, but it's a clean enough failure that I'm not going to worry about it.

### DIFF
--- a/src/internal/repository-metadata-files/provider/build.go
+++ b/src/internal/repository-metadata-files/provider/build.go
@@ -4,13 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"golang.org/x/mod/semver"
 	"log"
 	"os"
+	"slices"
+
+	"golang.org/x/mod/semver"
 	"registry-stable/internal"
 	"registry-stable/internal/github"
 	"registry-stable/internal/provider"
-	"slices"
 )
 
 func filterNewReleases(releases []github.GHRelease, existingMetadata provider.MetadataFile) ([]github.GHRelease, error) {


src/internal/github/git_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func Test_parseTagsFromStdout(t *testing.T) {
+	cases := map[string]struct {
+		input       []string
+		expected    []string
+		expectedErr string
+	}{
```
kislerdm, line 14:
```suggestion
	cases := map[string]struct {
		input       []string
		expected    []string
		expectedErr errror
	}{
```
...

src/internal/github/git_test.go
```diff
@@ -0,0 +1,67 @@
...
+				if err != nil {
+					t.Errorf("Unexpected error: %v", err)
+				}
+			} else {
+				if err == nil {
+					t.Errorf("Expected error %s, got <nil>", tc.expectedErr)
+				} else if err.Error() != tc.expectedErr {
+					t.Errorf("Expected error %s, got %v", tc.expectedErr, err)
+				}
+			}
```
kislerdm, line 59: I find this is easier to digest given that we use testify. WDYT?

```suggestion
if assert.ErrorIs(t, err, tc.expectedErr) {
    t.Fatalf("Unexpected error. Want: %v, got: %v", tc.expectedErr, err)
}
```

src/go.mod
```diff
@@ -3,6 +3,7 @@ module registry-stable
 go 1.21
 
 require (
+	github.com/google/go-cmp v0.5.9
```
kislerdm, line 6: **(nitpick)** I feel that we could use either cmp, or testify, both feels overhead. WDYT?

kislerdm (approved):
I left a couple of minor nitpicks, LGTM otherwise 👍🏻

src/internal/github/git.go
```diff
@@ -8,41 +8,55 @@ import (
...
-	for _, line := range strings.Split(buf.String(), "\n") {
+	for _, line := range lines {
 		if !strings.Contains(line, "refs/tags/") {
 			continue
 		}
 
 		fields := strings.Fields(line)
 		if len(fields) != 2 {
-			return nil, fmt.Errorf("could not parse tags for %s: tags are in wrong format", repositoryUrl)
+			return nil, fmt.Errorf("Invalid format for tag '%s', expected two fields", line)
```
cube2222, line 22: Generally, errors in Go shouldn't start with capital letters.

src/internal/github/git.go
```diff
@@ -8,41 +8,55 @@ import (
...
-	cmd := exec.Command("git", "ls-remote", "--tags", "--refs", repositoryUrl)
-	cmd.Stdout = &buf
-	cmd.Stderr = &bufErr
-
-	if err := cmd.Run(); err != nil {
-		log.Printf("Could not get tags for repository %s: %s", repositoryUrl, bufErr.String())
-		return nil, err
-	}
+func parseTagsFromStdout(lines []string) ([]string, error) {
+	exec.LookPath("")
```
Yantrio, line 12: What's this LookPath call doing here? 👀

src/internal/github/git.go
```diff
@@ -8,41 +8,55 @@ import (
...
+	log.Printf("Getting tags for repository %s", repositoryUrl)
+
+	var buf bytes.Buffer
+	var bufErr bytes.Buffer
+	cmd := exec.Command("git", "ls-remote", "--tags", "--refs", repositoryUrl)
+	cmd.Stdout = &buf
+	cmd.Stderr = &bufErr
+
+	if err := cmd.Run(); err != nil {
+		log.Printf("Could not get tags for repository %s: %s", repositoryUrl, bufErr.String())
```
Yantrio, line 51: Maybe this error should be a bit more specific, mainly around how executing the `git` command failed.

Maybe also reporting the exit statuscode too?

src/internal/github/git_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_parseTagsFromStdout(t *testing.T) {
+	cases := map[string]struct {
+		input       []string
+		expected    []string
+		expectedErr error
+	}{
+		"Simple Tag": {
```
Yantrio, line 16: It would be cool if we could add tests for empty string inputs (for fresh repos)

src/internal/github/git.go
```diff
@@ -8,41 +8,55 @@ import (
...
+		}
+
 		tags = append(tags, tag)
 	}
 
-	log.Printf("Found %d tags for repository %s", len(tags), repositoryUrl)
+	return tags, nil
+}
 
+func GetTags(repositoryUrl string) ([]string, error) {
```
Yantrio, line 41: Preferably I'd like GoDoc comments for atleast each exported function from a package. Even if it seems self explanatory, every little helps new developers approaching this.

Yantrio (changes requested):
Just a few nitpicks :) Otherwise it looks good, and I love the addition of the tests. Thanks!

src/internal/github/git.go
cam72cam: Updated to:
```
Could not get tags for www.creedthoughts.gov.wwwcreedthoughts: git failed (128) fatal: 'www.creedthoughts.gov.wwwcreedthoughts' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
```

Yantrio (approved):

PR #10, opened 2023-11-15
Unify github http client / auth

cam72cam:
This allows the same http client w/ oauth to be used in both the githubv4 API and rss requests.  It also moves the knowledge about the auth env var to a single function.


src/internal/github/client.go
```diff
@@ -16,3 +29,12 @@ func getGithubOauth2Client(ctx context.Context, token string) *http.Client {
 		&oauth2.Token{AccessToken: token},
 	))
 }
+
+func GetHttpRetryClient(ctx context.Context, token string) *http.Client {
```
kislerdm, line 33: **(nitpick)** Go-style 😄 

```suggestion
func GetHTTPRetryClient(ctx context.Context, token string) *http.Client {
```

kislerdm (approved):
I left a non-blocking tiny nit, otherwise LTGM. Thanks! 👍🏻

cam72cam:
I'm rolling this into another bit of work instead.

PR #11, opened 2023-11-15
Common json file output w/ tests

cam72cam:
I noticed a lot of copy pasta around writing to json files.  This changes @Yantrio's code to use @RLRabinowitz's files package, which has been enhanced with better error messages and tests.

### DIFF
--- a/src/internal/files/files_test.go
+++ b/src/internal/files/files_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestFiles_WriteToJsonFile_Success(t *testing.T) {
+func TestFiles_SafeWriteObjectToJsonFile_Success(t *testing.T) {
 	dir := t.TempDir()
 
 	data := map[string]interface{}{
@@ -19,7 +19,7 @@ func TestFiles_WriteToJsonFile_Success(t *testing.T) {
 
 	path := filepath.Join(dir, "subdir", "file.json")
 
-	err := WriteToJsonFile(path, data)
+	err := SafeWriteObjectToJsonFile(path, data)
 
 	if err != nil {
 		t.Error(err)
@@ -39,12 +39,12 @@ func TestFiles_WriteToJsonFile_Success(t *testing.T) {
 	assert.Equal(t, data, read)
 }
 
-func TestFiles_WriteToJsonFile_InvalidMarshall(t *testing.T) {
+func TestFiles_SafeWriteObjectToJsonFile_InvalidMarshall(t *testing.T) {
 	dir := t.TempDir()
 
 	path := filepath.Join(dir, "subdir", "file.json")
 
-	err := WriteToJsonFile(path, make(chan int))
+	err := SafeWriteObjectToJsonFile(path, make(chan int))
 
 	if err == nil {
 		t.Fatal("Expected marshal error, got <nil>")
@@ -52,11 +52,11 @@ func TestFiles_WriteToJsonFile_InvalidMarshall(t *testing.T) {
 	assert.Contains(t, err.Error(), "json: unsupported type: chan int")
 }
 
-func TestFiles_WriteToJsonFile_InvalidPath(t *testing.T) {
+func TestFiles_SafeWriteObjectToJsonFile_InvalidPath(t *testing.T) {
 	// TODO this might not be valid for non-posix systems
 	path := "/dev/null/foo"
 
-	err := WriteToJsonFile(path, nil)
+	err := SafeWriteObjectToJsonFile(path, nil)
 
 	if err == nil {
 		t.Fatal("Expected directory error, got <nil>")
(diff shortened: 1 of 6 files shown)


src/internal/files/files_test.go
```diff
@@ -0,0 +1,81 @@
...
+	path := filepath.Join(dir, "subdir", "file.json")
+
+	err := WriteToJsonFile(path, make(chan int))
+
+	if err == nil {
+		t.Error("Expected marshal error, got <nil>")
+	} else if !strings.Contains(err.Error(), "json: unsupported type: chan int") {
+		t.Errorf("Expected marshal error, got %s", err.Error())
+	}
+}
```
kislerdm, line 56: **(nitpick)** I'd rather suggest to use cmp, or testify for such comparisons since we have them in go.mod.

src/internal/v1api/modules.go
```diff
@@ -98,24 +83,8 @@ func (g Generator) writeModuleVersionDownload(namespace string, name string, sys
 	// the file should just contain a link to GitHub to download the tarball, ie:
 	// git::https://github.com/terraform-aws-modules/terraform-aws-iam?ref=v5.30.0
 	location := fmt.Sprintf("git::github.com/%s/terraform-%s-%s?ref=%s", namespace, name, system, version)
```
kislerdm, line 85: **(issue)**

```suggestion
location := fmt.Sprintf("git::github.com/%s/terraform-%s-%s?ref=%s", namespace, system, name, version)
```

kislerdm (approved):
LGTM, created #12 as a followup.

src/internal/v1api/modules.go
```diff
@@ -72,24 +73,8 @@ func (g Generator) readModuleMetadata(path string, logger *slog.Logger) (*module
...
-	}
-
-	err = os.WriteFile(filePath, marshalled, 0644)
-	if err != nil {
-		return fmt.Errorf("failed to write file: %w", err)
-	}
-
-	return nil
+	path := filepath.Join(g.DestinationDir, "v1", "modules", namespace, name, targetSystem, "versions")
+	return files.WriteToJsonFile(path, ModuleVersionListingResponse{Modules: []ModuleVersionListingResponseItem{{Versions: versions}}})
```
Yantrio, line 77: It's a small nitpick on my side and it may seem like I'm being very hesitant unnecessarily, but I feel like "magic" functions that do more than what they're named are a bit of a code-smell and I would prefer to be explicit with the multiple stages here.

I'm happy to have a method to wrap making directories and writing to files, but I think using one single method to do both is a bit too "magic" for what is needed.

either that, or we need to rename the method to be something quite verbose to explain what it's doing.

As a developer approaching this code, I may question "Where is the directory made?" and not see this at a glance without having to dive in deep with how we're writing files to the disk. 

This is why I did not use this existing method in the files package. Infact I, personally, would like to remove the files package as I feel like it's not needed. 

However, I welcome others to propose alternative points of view

Yantrio, line 77: If we are keeping this method call the way it is, I would rather explicitly handle the error here too instead of returning directly. 

this allows us to wrap the error nicely and provide information about what failed and where.

src/internal/files/files_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFiles_WriteToJsonFile_Success(t *testing.T) {
```
Yantrio, line 13: do we really need unit tests for a method that 100% just wraps standard lib?

I feel like we're not testing any of our code here, we're testing things that already exist and are maintained elsewhere.

cam72cam: For me, it helped shape the error messages.  It's definitely a bit overkill, but IMO probably does not hurt?

src/internal/v1api/modules.go
cube2222: I personally feel like this function is useful.

In practice the pattern of "serialize to a formatted json, make sure directory exists, write to file in that directory" will happen in quite a few places in the registry codebase.

No strong opinion about the packaging or function name, but the idea for the function itself, to me, is sound. And testing it makes sense to me as well (to make sure the function works when the directory doesn't exist, etc.).

I don't see value in repeating this code in every place it's needed.

Yantrio: I'm happy to accept the value of this method, but then we should attempt to name it in a clear manner so that it's understandable to someone consuming this code what it does.

To keep it short im happy with:
- SafeDumpToJsonFile
- SafeWriteObjectToJsonFile

or something similar, this makes it a bit clearer that it's doing a little bit more than os.WriteFile under the hood.

cam72cam: `SafeWriteObjectToJsonFile` works for me

cam72cam: Changed in f69a096

src/internal/files/files_test.go
Yantrio: I think if we want to test the error messages then that's fine :D

Yantrio (approved):

PR #14, opened 2023-11-15
Define the generator / module package responsibilities

cam72cam:
I'm looking for feedback on moving some of the generator logic closer to the Module struct.

### DIFF
--- a/src/internal/files/files.go
+++ b/src/internal/files/files.go
@@ -11,7 +11,6 @@ func SafeWriteObjectToJsonFile(filePath string, data interface{}) error {
 	marshalledJson, err := json.MarshalIndent(data, "", "  ")
 	if err != nil {
 		return fmt.Errorf("failed to marshal for %s: %w", filePath, err)
-
 	}
 
 	err = os.MkdirAll(path.Dir(filePath), 0755)


src/cmd/generate-v1/main.go
```diff
@@ -27,6 +27,7 @@ func main() {
 
 		ModuleDirectory:   *moduleDataDir,
 		ProviderDirectory: *providerDataDir,
+		Logger:            slog.Default(),
```
Yantrio, line 30: What's the reason for having the logger as part of the owning type?

I feel like there are cases where we are immediately going to create a new logger when we call `logger.With` at the top level exported methods of this package.

src/internal/module/module.go
```diff
@@ -16,7 +21,32 @@ type Module struct {
 	TargetSystem string // The module target system
 }
 
+func (m Module) Logger(logger *slog.Logger) *slog.Logger {
+	return logger.With(slog.String("namespace", m.Namespace), slog.String("name", m.Name), slog.String("targetSystem", m.TargetSystem))
```
Yantrio, line 25: did we need a method for one line?

src/internal/module/module.go
```diff
@@ -16,7 +21,32 @@ type Module struct {
...
+}
```
Yantrio, line 26: If we're going to have a method to enrich the logger we should explain what it does in the method name and also do it in a bit of a more maintainable way rather than just one line.

For example: https://github.com/opentofu/registry/blob/b148dfce0206c19baca80a1d4c137b97722067aa/src/lambda/api/moduleDownload.go#L24-L32

Also, if this is returning a logger, maybe something like `m.EnrichLogger(logger *slog.Logger)` would be good as that explains what it's doing.


If I didn't read the contents of this method, I would be confused why I would have to call: `logger := m.Logger(slog.Default())`, but I would understand why if it was something like `logger := m.EnrichLogger(slog.Default())`

src/internal/module/module.go
```diff
@@ -16,7 +21,32 @@ type Module struct {
...
+
 func (m Module) RepositoryURL() string {
-	repoName := fmt.Sprintf("terraform-%s-%s", m.TargetSystem, m.Name)
-	return fmt.Sprintf("https://github.com/%s/%s", m.Namespace, repoName)
+	return fmt.Sprintf("https://github.com/%s/terraform-%s-%s", m.Namespace, m.TargetSystem, m.Name)
```
Yantrio, line 29: We should be careful here, if we ever want to introduce a proxy or support anything else we may not want to hardcode github.com here.

What I would recommend is that we have a method on the github client where you can pass it an org/repo and it will generate you the url.

src/internal/module/module_test.go
```diff
@@ -0,0 +1,35 @@
...
+	assert.Equal(t, "dir/v1/modules/spacename/name/target/1.0.1/download", m.VersionDownloadPath("./dir", v))
+
+	buff := bytes.NewBufferString("")
+	bw := bufio.NewWriter(buff)
+
+	logger := slog.New(slog.NewJSONHandler(bw, nil))
+	m.Logger(logger).Info("Module")
+	bw.Flush()
+
+	assert.Contains(t, buff.String(), `"namespace":"spacename","name":"name","targetSystem":"target"`)
```
Yantrio, line 34: I don't think we need to test slog here. 
```suggestion

```

src/internal/v1api/modules.go
```diff
@@ -17,41 +15,44 @@ import (
 // For more information see
 // https://opentofu.org/docs/internals/module-registry-protocol/#list-available-versions-for-a-specific-module
 // https://opentofu.org/docs/internals/module-registry-protocol/#download-source-code-for-a-specific-module-version
-func (g Generator) GenerateModuleResponses(_ context.Context, namespace string, name string, targetSystem string) error {
-	logger := slog.With(slog.String("namespace", namespace), slog.String("name", name), slog.String("targetSystem", targetSystem))
+func (g Generator) GenerateModuleResponses(_ context.Context, m module.Module) error {
```
Yantrio, line 18: Whilst I understand the simplicity of passing in the module here, I kept things seperate as then the caller of the method does not have to construct an object to call a method on it's own.

If we're looking at this package as an individual component, rather than part of the whole then I would rather reach for the ease of calling, rather than tying it to the calling of these methods to the module package.

src/internal/v1api/modules.go
```diff
@@ -65,26 +66,25 @@ func (g Generator) readModuleMetadata(path string, logger *slog.Logger) (*module
...
 }
 
 // writeModuleVersionListing writes the file containing the module version listing.
 // This data  is to be consumed when an end user requests /v1/modules/{namespace}/{name}/{targetSystem}/versions
-func (g Generator) writeModuleVersionListing(namespace string, name string, targetSystem string, versions []ModuleVersionResponseItem) error {
-	path := filepath.Join(g.DestinationDir, "v1", "modules", namespace, name, targetSystem, "versions")
-	return files.WriteToJsonFile(path, ModuleVersionListingResponse{Modules: []ModuleVersionListingResponseItem{{Versions: versions}}})
+func (g Generator) writeModuleVersionListing(m module.Module, versions []ModuleVersionResponseItem) error {
+	return files.WriteToJsonFile(
+		m.VersionListingPath(g.DestinationDir),
```
Yantrio, line 78: I don't see why the module package has to understand anything about the destination directory.

Maybe I would prefer something like `filepath.Join(g.DestinationDir, m.VersionListingPath)

cam72cam:
James and I spoke, next step is to pull out the logging changes into it's own changeset

src/internal/v1api/modules.go
cam72cam: The goal was to remove additional calls to join path, but it's not a performance issue so I'll adopt your suggestion

Yantrio:
I've spoken to @cam72cam about this PR. I'm happy with the changes overall now that the logging has been pulled out. We will discuss logging changes overall once the registry is a bit more complete.

Yantrio:
After discussion with @cam72cam , I agree that tying these 2 areas of code together does make sense. Mainly because I do not expect anyone to be consuming the generator without the metadata in place.

For this reason, it's a bit nicer for devs to pass around objects rather than trying to decouple things.

Yantrio (approved):

cube2222 (approved):

src/internal/module/module.go
```diff
@@ -17,6 +21,27 @@ type Module struct {
...
 
 func (m Module) RepositoryURL() string {
-	repoName := fmt.Sprintf("terraform-%s-%s", m.TargetSystem, m.Name)
-	return fmt.Sprintf("https://github.com/%s/%s", m.Namespace, repoName)
+	return fmt.Sprintf("https://github.com/%s/terraform-%s-%s", m.Namespace, m.TargetSystem, m.Name)
+}
+
+// the file should just contain a link to GitHub to download the tarball, ie:
+// git::https://github.com/terraform-aws-modules/terraform-aws-iam?ref=v5.30.0
+func (m Module) VersionDownloadURL(version Version) string {
```
kislerdm, line 29: **(todo)** to add tests for the methods -> @kislerdm

src/internal/v1api/modules.go
```diff
@@ -18,41 +16,49 @@ import (
 // For more information see
 // https://opentofu.org/docs/internals/module-registry-protocol/#list-available-versions-for-a-specific-module
 // https://opentofu.org/docs/internals/module-registry-protocol/#download-source-code-for-a-specific-module-version
-func (g Generator) GenerateModuleResponses(_ context.Context, namespace string, name string, targetSystem string) error {
-	logger := slog.With(slog.String("namespace", namespace), slog.String("name", name), slog.String("targetSystem", targetSystem))
+func (g Generator) GenerateModuleResponses(_ context.Context, m module.Module) error {
+	logger := slog.With(slog.String("namespace", m.Namespace), slog.String("name", m.Name), slog.String("targetSystem", m.TargetSystem))
```
kislerdm, line 20: **(idea)** We may want to refactor it out as a Generator attribute/method to facilitate/unify logging/observability.

kislerdm (approved):
LGTM! 👍🏻 

For context, the comments I left are for the sake of documentation - to revisit them and create issue.

PR #16, opened 2023-11-16
Add command to bump provider and module versions

RLRabinowitz:
- Create new `main.go`, adds a command to version bump all existing providers and modules
- Required propagation of the module and provider directory to many methods (we'd probably want to change it later to be similar to `Generator`
- Added `list.go` to list all existing providers
- Add an actual provider to the registry

### DIFF
--- a/src/cmd/bump-versions/main.go
+++ b/src/cmd/bump-versions/main.go
@@ -12,8 +12,8 @@ func main() {
 	slog.SetDefault(slog.New(slog.NewJSONHandler(os.Stdout, nil)))
 	slog.Info("Starting version bump process for modules and providers")
 
-	moduleDataDir := flag.String("module-data", "../../../modules", "Directory containing the module data")
-	providerDataDir := flag.String("provider-data", "../../../providers", "Directory containing the provider data")
+	moduleDataDir := flag.String("module-data", "../modules", "Directory containing the module data")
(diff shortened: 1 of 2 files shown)


src/internal/repository-metadata-files/provider/create.go
```diff
@@ -1,22 +1,21 @@
...
-	filePath := getFilePath(p)
+	filePath := getFilePath(p, providerDataDir)
 	return files.WriteToFile(filePath, repositoryFileData)
 }
 
-func getFilePath(p provider.Provider) string {
-	shard := p.Namespace[0]
-	return fmt.Sprintf("providers/%c/%s/%s.json", shard, p.Namespace, p.ProviderName)
+func getFilePath(p provider.Provider, providerDataDir string) string {
+	return filepath.Join(providerDataDir, p.Namespace[0:1], p.Namespace, p.ProviderName+".json")
```
cube2222, line 20: Should the shard name be case-insensitive? I think so.

Basically, I'd use strings.ToLower for the shard.

cube2222 (approved):

src/internal/repository-metadata-files/provider/create.go
RLRabinowitz: Done in e8f075dab560789e332c10e33a63c697867bcb74

We'd need to later refactor how we get the module/provider directory path

Yantrio (approved):

PR #21, opened 2023-11-21
Add all OpenTofu providers to the registry

RLRabinowitz:
I've added all existing OpenTofu providers (by running `CreateMetadataFile` on all of them). I've also removed `ultradns`, which should not really be a published provider

I've also added the providers to exist under the `hashicorp` namespace


cam72cam (approved):
I'm planning on regenerating all of these once https://github.com/opentofu/registry-stable/pull/22 is merged.  The semver comparison was invalid and ordered from oldest to newest.  Specific changes: https://github.com/opentofu/registry-stable/pull/22/commits/68594bca1f175dbd7ecdcd07778920fd06efd0c0

kislerdm (approved):

Yantrio:
If we need to regenerate after 22 is merged, it may be better to get #22 in first, then we don't have to do a lot of manual cleanup before we have a re-generation step in place.

I'll also raise an issue for us to track in the future that we want to be able to invalidate or even remove some items without manual intervention in the r2 bucket.

cam72cam:
I don't mind either way @Yantrio, for now I just run:
```
find ../providers/ | grep 'json$' | xargs -I foo bash -c "echo '{}' > foo"
~/go/bin/bump-versions
```

It takes 5-7 minutes.

PR #22, opened 2023-11-21
Provider + github refactoring

cam72cam:
Goals:
* Clear split between providers definitions  vs generated providers
  - src/internal/v1api/source.go maps from definitions to generated objects
* Unify object oriented(ish) approach to providers
* Use above to keep track of logger/ctx/github clients
* Clean up parallelization / rate limiting
* Improve error messages
* Improve logging

I've tested this with all of the providers in #21.
* Time with unmodified providers: 4s
* Time with empty providers {}: 6 min for full regen from github

### DIFF
--- a/src/internal/github/client.go
+++ b/src/internal/github/client.go
@@ -36,7 +36,7 @@ func NewClient(ctx context.Context, log *slog.Logger, token string) Client {
 	}}
 	httpClientFast := &http.Client{Transport: &transport{
 		ctx:         ctx,
-		ratelimiter: rate.NewLimiter(rate.Every(time.Second/60), 1),
+		ratelimiter: rate.NewLimiter(rate.Every(time.Second/30), 1),
 	}}
 
 	return Client{


src/cmd/bump-versions/main.go
```diff
@@ -1,31 +1,36 @@
...
+	"registry-stable/internal/provider"
 	"registry-stable/internal/repository-metadata-files/module"
-	"registry-stable/internal/repository-metadata-files/provider"
 )
 
 func main() {
-	slog.SetDefault(slog.New(slog.NewJSONHandler(os.Stdout, nil)))
-	slog.Info("Starting version bump process for modules and providers")
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
+	slog.SetDefault(logger) // TODO REMOVE ME
```
kislerdm, line 15: **(question)** Why not remove within the scope of this PR's changes? 🤔

cam72cam: I was trying to leave the module path alone, which currently uses the log/slog default logger (I think).

src/cmd/bump-versions/main.go
```diff
@@ -38,13 +43,34 @@ func main() {
...
+
+	errChan := make(chan error, len(providers))
+
 	for _, p := range providers {
-		slog.Info("Beginning version bump process for provider", slog.String("provider", p.Namespace+"/"+p.ProviderName))
-		err = provider.UpdateMetadataFile(p, *providerDataDir)
+		p := p
+		go func() {
+			errChan <- p.UpdateMetadataFile()
+		}()
```
kislerdm, line 58: **(todo)** We may want to execute goroutines in batches to avoid running 10k+ goroutines concurrently, especially given that our process will be given 2 vCPU (4 cores/"Machines" for goroutines at most). Overall, concurrency feels a bit preliminary at this point. Did we hit performance bottleneck already? WDYT?

cam72cam: Most of them will quickly hit the http rate limiter baked into the github client and be in a low overhead wait state.

cam72cam: For a full refresh, it looks like it uses a few % cpu on a single core.

src/cmd/bump-versions/main.go
kislerdm: Thanks for clarification!

src/internal/github/client.go
```diff
@@ -2,17 +2,77 @@ package github
...
+
+func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	err := t.ratelimiter.Wait(t.ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	req = req.WithContext(t.ctx)
+	req.Header.Set("User-Agent", "OpenTofu Registry/1.0")
+	req.Header.Set("Authorization", "Bearer "+t.token)
```
kislerdm, line 76: **(suggestion)** Since we don't validate that token is a non-empty string:

```suggestion
        if t.token != "" {
            req.Header.Set("Authorization", "Bearer "+t.token)
        }
```

src/internal/github/client.go
```diff
@@ -2,17 +2,77 @@ package github
...
+	log        *slog.Logger
+	httpClient *http.Client
+	ghClient   *githubv4.Client
+}
+
+func NewClient(ctx context.Context, log *slog.Logger, token string) Client {
+	httpClient := &http.Client{Transport: &transport{
+		token:       token,
+		ctx:         ctx,
+		ratelimiter: rate.NewLimiter(rate.Every(time.Second/25), 1),
```
kislerdm, line 35: **(question)** Could you please share the motivation behind those numbers? I'm not sure I can reasonably guess it 😃 Thanks!

src/internal/github/rss.go
```diff
@@ -42,37 +40,15 @@ func extractTag(item *gofeed.Item) (string, error) {
...
-				err = ce
-			}
-		}()
+		return nil, fmt.Errorf("%s: %w", releasesRssUrl, err)
 	}
+	defer resp.Body.Close()
 
-	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		return nil, fmt.Errorf("got error %d", resp.StatusCode)
+	if resp.StatusCode != http.StatusOK {
```
kislerdm, line 50: **(idea)** WDYT about covering 429 with a backoff/retry pattern?

kislerdm (commented):
Hey Christian! Well done, thank you for the effort! 🚀 

I left a few non-blocking comments, I'd be happy to discuss them in sync if you'd prefer. Also, I may give the changes another read-through tomorrow. Although, please feel free to merge PR - my comments are not blocking. Thanks!

src/cmd/bump-versions/main.go
```diff
@@ -38,13 +43,34 @@ func main() {
...
-			slog.Error("Failed to version bump provider", slog.Any("err", err))
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) != 0 {
+		logger.Error("Encountered errors while processing providers")
+		for _, err := range errs {
+			logger.Error(err.Error())
 		}
 	}
```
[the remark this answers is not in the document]
cam72cam: There are two levels of concurrency at play here:
* Each provider looking up versions it might need to fetch (rss/githubv4)
  - Each version that needs to be fetched making 10s to 100s of HTTP reqs for shasums/metadata

Each of those is throttled inside the shared github.Client object so that it can't overwhelm the endpoints, even when new functionality / routines are added.

I'd wait to optimize this further until we start hitting go-routine bottlenecks:
[It is practical to create hundreds of thousands of goroutines in the same address space](https://go.dev/doc/faq#goroutines)

src/internal/github/client.go
```diff
@@ -2,17 +2,77 @@ package github
...
+}
+
+type Client struct {
+	ctx        context.Context
+	log        *slog.Logger
+	httpClient *http.Client
+	ghClient   *githubv4.Client
+}
+
+func NewClient(ctx context.Context, log *slog.Logger, token string) Client {
```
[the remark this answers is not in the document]
cam72cam: I'd rather move github.Client to an interface when we get to that level of testing.  That would allow us to easily create a mock that would implement the github.Client interface without hacking an existing struct.

src/internal/github/client.go
cam72cam: It is an expectation of this API that token is not an empty string (checked for in the one place that it's currently read).  With an application this small, I think this is an unnecessary check.

src/internal/github/rss.go
cam72cam: That's a good thought, though it would be neat to do that inside of the backing httpClient transparently.  There's some commented out code in github.NewClient that had a similar idea.  Github's API theoretically has some good response headers that can tell us if we are hitting it too hard.

I think this is an ongoing discussion and might be worth creating an issue in the repo?

src/cmd/bump-versions/main.go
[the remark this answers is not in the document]
cube2222: Also, if you decide to limit concurrency, please use a buffered-channel-based semaphore.

Basically
```
sem := make(chan struct{}, maxParallelism)

// before spawning goroutine
sem <- struct{}{}

// in goroutine defer
<- sem
```

src/internal/provider/protocols.go
```diff
@@ -14,14 +12,12 @@ type ManifestMetadata struct {
...
 
-func GetProtocols(ctx context.Context, manifestDownloadUrl string) ([]string, error) {
-	if manifestDownloadUrl == "" {
-		slog.Warn("Could not find manifest file, using default protocols")
-		return []string{"5.0"}, nil
+func (p Provider) GetProtocols(manifestDownloadUrl string) ([]string, error) {
+	contents, err := p.Github.DownloadAssetContents(manifestDownloadUrl)
+	if err != nil {
+		return nil, err
 	}
```
Yantrio, line 20: It's often the case that the manifest doesnt exist, and in these cases we should assume that it is 5.0, 

can you add this logic back in please?

src/internal/github/assets.go
```diff
@@ -18,51 +15,25 @@ type Platform struct {
 	Arch string
 }
 
-func DownloadAssetContents(ctx context.Context, downloadURL string) ([]byte, error) {
-	httpClient := httpInternal.GetHttpRetryClient()
-
-	log.Printf("Downloading asset, url: %s", downloadURL)
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, downloadURL, nil)
+func (c Client) DownloadAssetContents(downloadURL string) ([]byte, error) {
```
Yantrio, line 18: I still think we should pass a context here so we can do context propogation and timeouts if we want to later on.

Then we could have the http client use that context too.

src/internal/provider/version.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+			target.DownloadURL, ok = assets[target.Filename]
+			if !ok {
+				// Artifact does not exist for this platform
+				continue
+			}
+
+			v.Targets = append(v.Targets, target)
+		}
+	}
```
Yantrio, line 57: Whilst I like this approach right now, it may become a pain if a new OS or architecture comes out and we have to remember to update the registry to support it too. 

I think the old way of inferring the arch/os from the filename was a bit better at handling this situation. A perfect example is the recent-ish release of arm on darwin

src/internal/v1api/source.go
```diff
@@ -0,0 +1,64 @@
...
+				SHASumsURL:          ver.SHASumsURL,
+				SHASumsSignatureURL: ver.SHASumsSignatureURL,
+				SHASum:              target.SHASum,
+				SigningKeys:         SigningKeys{}, // TODO: Add gpg keys
+			}
+			versionDetails[p.VersionDownloadPath(ver, details)] = details
+		}
+	}
+	return versionDetails
+}
```
Yantrio, line 64: I love this 👍 Maybe once the module stuff is in we can either duplicate or have one central "metadatasource" that we can use for both provider and module data.

src/internal/provider/protocols.go
cam72cam: I moved this logic to the caller, it's not removed.  IMO it's cleaner to have it in there, that's where we validate / parse the release information.

src/internal/v1api/source.go
cam72cam: Excellent!

src/internal/provider/protocols.go
Yantrio: I see this is moved there so I'm happy with that 👍

src/internal/github/assets.go
Yantrio: Context is handled in the Client, and the client should only live as long as it needs, so im fine with this. especially when it helps with context cancellation

PR #32, opened 2023-11-22
Refactor common CLI operations

cam72cam:
Reduces the copy/pasta between commands.  Also speeds up the generator 4-5x.

### DIFF
--- a/src/cmd/bump-versions/main.go
+++ b/src/cmd/bump-versions/main.go
@@ -2,22 +2,18 @@ package main
 
 import (
 	"context"
-	"flag"
 	"log/slog"
 	"os"
+	"registry-stable/cmd/common"
 	"registry-stable/internal/github"
 	"registry-stable/internal/module"
 	"registry-stable/internal/provider"
 )
 
 func main() {
-	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
-	logger.Info("Starting version bump process for modules and providers")
+	cli := common.Parse()
 
-	moduleDataDir := flag.String("module-data", "../modules", "Directory containing the module data")
-	providerDataDir := flag.String("provider-data", "../providers", "Directory containing the provider data")
-
-	flag.Parse()
+	cli.Logger.Info("Starting version bump process for modules and providers")
 
 	ctx := context.Background()
 	token, err := github.EnvAuthToken()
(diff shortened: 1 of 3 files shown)


src/cmd/bump-versions/main.go
```diff
@@ -2,88 +2,38 @@ package main
...
+	"registry-stable/cmd/common"
 	"registry-stable/internal/github"
 	"registry-stable/internal/module"
 	"registry-stable/internal/provider"
 )
 
 func main() {
-	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
-	logger.Info("Starting version bump process for modules and providers")
+	cli := common.Parse()
```
kislerdm, line 14: **(idea/nitpick)** WDYT about the package `config` with the function `Configure()` instead? Motivation: the `common` package does not only parse CLI arguments, but also configures the app (e.g. defines the logger) following the builder pattern.

kislerdm (approved):
LGTM 👍🏻 Also, I left a non-blocking idea as comment 😄

src/cmd/bump-versions/main.go
Yantrio: I agree, I really don't like packages called things like "utils" or "common"

`config` would be good I think.

Yantrio: Also it should go into `internal` instead of `cmd`, otherwise people will try and run `go run ./cmd/common`

src/cmd/common/common.go
```diff
@@ -0,0 +1,85 @@
...
+type CLI struct {
+	Logger          *slog.Logger
+	Github          github.Client
+	ModuleDataDir   string
+	ProviderDataDir string
+}
+
+func Parse() CLI {
+	moduleDataDir := flag.String("module-data", "../modules", "Directory containing the module data")
+	providerDataDir := flag.String("provider-data", "../providers", "Directory containing the provider data")
```
Yantrio, line 22: Whilst these 2 flags are common between the 2 commands right now, they may not be common across everything we have planned in the future. I think this is a bit of an over-optimization that isn't yet needed. However I'm happy to discuss this more.

src/cmd/common/common.go
```diff
@@ -0,0 +1,85 @@
...
+	cli := CLI{
+		Logger:          slog.New(slog.NewJSONHandler(os.Stdout, nil)),
+		ModuleDataDir:   *moduleDataDir,
+		ProviderDataDir: *providerDataDir,
+	}
+
+	return cli
+}
+
+func (cli CLI) Modules(action func(m module.Module) error) error {
```
Yantrio, line 35: Whilst we have 2 commands now that iterate over modules and providers, I feel like this method isn't really needed. Especially as not every CLI tool we make is going to use this.

For example, We have 3-5 tools planned to look at just GPG keys, and they wont use these at all.

Yantrio: I wouldn't be against a method similar to this in the module/provider packages though. Something like

`modules.ForEachModule(modulePath string, action func(m module.Module)....`

Yantrio (changes requested):
Overall I'm not a big fan of this, as i feel this may be an early optimisation and i think we should re-think this once we have the rest of the cli tools in the repo.

src/cmd/common/common.go
kislerdm: I'd also say that it shall be a single flag `"--output"` and the cmd's are rather split on the cmd level:
```commandline
go run ./cmd/bumper/provider --output=/tmp/foo
go run ./cmd/bumper/module --output=/tmp/bar
```

alternatively:

```commandline
go run ./cmd/bumper provider --output=/tmp/foo
go run ./cmd/bumper module --output=/tmp/foo
```

src/cmd/bump-versions/main.go
kislerdm: > I agree, I really don't like packages called things like "utils" or "common"

I'd be a proponent of the team agreement to _not_ use those generic names as they have the "hide under the carpet" effect and carry not much meaning.

cam72cam:
Alright, I'll close this and open up a simpler one with the concurrency helpers moved into internal.  The main motivation was cleaning up a lot of the duplicate parallel code which had a few bugs from the copy/paste.

PR #33, opened 2023-11-23
Refactor cli parallelism to reduce copy/paste errors

cam72cam:
Based on comments in https://github.com/opentofu/registry-stable/pull/32, I created a simplified PR which makes a lot less choices that could limit future designs.

### DIFF
--- a/src/internal/parallel/each.go
+++ b/src/internal/parallel/each.go
@@ -1,0 +1,30 @@
+package parallel
+
+type Action func() error
+
+func ForEach(actions []Action, maxConcurrency int) []error {
+	// Populate tokens
+	tokens := make(chan int, maxConcurrency)
+	for i := 0; i < maxConcurrency; i++ {
+		tokens <- i
+	}
+
+	errChan := make(chan error, len(actions))
+	for _, a := range actions {
+		a := a
+		token := <-tokens
+		go func() {
+			defer func() { tokens <- token }()
+			errChan <- a()
+		}()
+	}
+
+	var errs []error
+	for _ = range actions {
+		err := <-errChan
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return errs
+}
--- a/src/cmd/bump-versions/main.go
+++ b/src/cmd/bump-versions/main.go
@@ -22,7 +22,7 @@ func main() {
 	ctx := context.Background()
 	token, err := github.EnvAuthToken()
 	if err != nil {
-		slog.Error("Initialization Error", slog.Any("err", err))
+		logger.Error("Initialization Error", slog.Any("err", err))
 		os.Exit(1)
 	}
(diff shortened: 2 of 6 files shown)


cam72cam:
Example output:
```json
...
{"time":"2023-11-23T08:20:47.087076879-05:00","level":"ERROR","msg":"https://github.com/oracle/terraform-provider-oci/releases.atom got error 429","type":"provider","provider":{"namespace":"oracle","name":"oci"}}
{"time":"2023-11-23T08:20:47.252799613-05:00","level":"ERROR","msg":"Encountered 29 errors processing 73 providers"}
```

Yantrio (approved):

src/internal/parallel/each.go
```diff
@@ -0,0 +1,30 @@
+package parallel
+
+type Action func() error
+
+func ForEach(actions []Action, maxConcurrency int) []error {
```
cube2222, line 5: I propose replacing both this function and the Provider/ModuleList.Parallel function with the following
```
import (
	"golang.org/x/sync/errgroup"
)

func ParallelForEach[T any](elements []T, maxConcurrency int, f func(index int, element T) error) error {
	sem := make(chan struct{}, maxConcurrency)
	var g errgroup.Group
	for i, element := range elements {
		sem <- struct{}{}
		i, element := i, element
		g.Go(func() error {
			defer func() { <-sem }()
			return f(i, element)
		})
	}
	return g.Wait()
}
```

cam72cam: I'd prefer to leave the Provider/ModuleList.Parallel functions as they are, since they also handle logging specific actions.  I'll take a look at refactoring those using your proposal.

cube2222: Sure

cam72cam: It also looks like errgroup allows you to set a max concurrency internally, dramatically simplifying the code!

cube2222: Ah, TIL!

cam72cam: Hmm, this also has the side effect of only returning the first error instead of a list of errors.  Is that preferred?

cube2222: I think that's fine, yeah.

cam72cam: If that's the desired behavior, it becomes more complex as you would want to call WithContext(...) on the group with the command's context.

cube2222: Alright, I don't mind either way. Whatever is simpler.

kislerdm (approved):

Yantrio (approved):

src/internal/parallel/each.go
cam72cam: https://github.com/opentofu/registry-stable/issues/40

PR #42, opened 2023-11-24
Updated module name, introduced linting, fixed a few errors of mis-called properties

Yantrio:
I will put in more PRs soon to add more comments and testing on top of this branch.

### DIFF
--- a/src/.golangci.yml
+++ b/src/.golangci.yml
@@ -1,0 +1,90 @@
+run:
+  timeout: 1m
+
+linters-settings:
+  errcheck:
+    check-type-assertions: true
+
+  exhaustive:
+    check:
+      - switch
+      - map
+
+  funlen:
+    lines: 100
+    statements: 50
+    ignore-comments: true
+
+  govet:
+    enable-all: true
+    disable:
+      - fieldalignment # rule is too strict
+
+  nolintlint:
+    require-explanation: true
+    require-specific: true
+
+linters:
+  disable-all: true
+  enable:
+    - asasalint
+    - asciicheck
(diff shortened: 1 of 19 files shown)


cube2222 (approved):

src/.golangci.yml
```diff
@@ -0,0 +1,90 @@
+run:
```
ghost, line 1: Shouldn't this rather be in the root directory? It would seem that `src` implies that there's source code there.

src/go.mod
```diff
@@ -1,4 +1,4 @@
-module registry-stable
+module github.com/opentofu/registry-stable
```
ghost, line 1: I would suggest creating a redirector subdomain for the module coordinates so we are flexible in terms of Git hosting. ([see this example](https://github.com/ContainerSSH/go.containerssh.io))

ghost (approved):

src/.golangci.yml
Yantrio: The root directory is for the files of the registry (metadata for the providers/modules), src is where all the source code is.

src/go.mod
Yantrio: That's a good idea. We should raise an issue to do something similar in the future. I'll raise one after my lunch

kislerdm (approved):

cam72cam (approved):

src/go.mod
Yantrio: I've created https://github.com/spacelift-io/opentofu-registry/issues/2 for this

PR #43, opened 2023-11-24
More cleanup of the codebase, ready for other contributors to work ontop of.

Yantrio:
This is the second half of #42 

In this PR I'm mainly:
- Improving some readability of code by re-ordering things or renaming things
- Introducing comments wherever I found them missing
- Adding some tests for easily testable areas of code

### DIFF
--- a/src/internal/files/files.go
+++ b/src/internal/files/files.go
@@ -7,9 +7,9 @@ import (
 	"path"
 )
 
-// SafeWriteObjectToJSONFile writes the given object to the given file path.
+// SafeWriteObjectToJSONFile writes the given data to the given file path.
 // It also ensures that the destination directory exists and that the file is written correctly.
-func SafeWriteObjectToJSONFile(filePath string, data interface{}) error {
+func SafeWriteObjectToJSONFile(filePath string, data any) error {
 	marshalledJSON, err := json.MarshalIndent(data, "", "  ")
 	if err != nil {
 		return fmt.Errorf("failed to marshal for %s: %w", filePath, err)
@@ -22,7 +22,7 @@ func SafeWriteObjectToJSONFile(filePath string, data interface{}) error {
 
 	err = os.WriteFile(filePath, marshalledJSON, 0600) //nolint: gomnd // 0600 is fine for what we need, no other users should consume this
 	if err != nil {
-		// Error already contains filePath
+		// Error already contains filePath so we don't need to add it again
 		return fmt.Errorf("failed to write to file: %w", err)
 	}
 
--- a/src/internal/github/client.go
+++ b/src/internal/github/client.go
@@ -11,14 +11,18 @@ import (
 	"github.com/shurcooL/githubv4"
 )
 
+const UserAgent = "OpenTofu Registry/1.0"
+
+// EnvAuthToken returns the GitHub token from the environment.
 func EnvAuthToken() (string, error) {
 	token := os.Getenv("GH_TOKEN")
 	if token == "" {
-		return "", fmt.Errorf("Expected $GH_TOKEN to be set, unable to authenticate with GitHub")
+		return "", fmt.Errorf("expected the GH_TOKEN environment variable to be set, unable to authenticate with GitHub")
 	}
 	return token, nil
 }
 
+// Client is a GitHub client that abstracts away the different GitHub APIs and handles rate limiting/throttling.
 type Client struct {
 	ctx        context.Context
 	log        *slog.Logger
@@ -31,6 +35,7 @@ type Client struct {
 	rssThrottle   Throttle
 }
 
+// NewClient creates a new GitHub client.
 func NewClient(ctx context.Context, log *slog.Logger, token string) Client {
 	httpClient := &http.Client{Transport: &transport{token: token, ctx: ctx}}
 	return Client{
@@ -52,6 +57,7 @@ func NewClient(ctx context.Context, log *slog.Logger, token string) Client {
 	*/
 }
 
+// WithLogger returns a new Client with the given logger.
 func (c Client) WithLogger(log *slog.Logger) Client {
 	return Client{
 		ctx:        c.ctx,
@@ -66,14 +72,17 @@ func (c Client) WithLogger(log *slog.Logger) Client {
 	}
 }
 
+// transport is a http.RoundTripper that makes sure all requests have the
+// correct User-Agent and Authorization headers set.
 type transport struct {
 	token string
 	ctx   context.Context
 }
 
+// RoundTrip is needed to implement the http.RoundTripper interface.
 func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	req = req.WithContext(t.ctx)
-	req.Header.Set("User-Agent", "OpenTofu Registry/1.0")
+	req.Header.Set("User-Agent", UserAgent)
 	req.Header.Set("Authorization", "Bearer "+t.token)
 	return http.DefaultTransport.RoundTrip(req)
 }
--- a/src/internal/github/rss.go
+++ b/src/internal/github/rss.go
@@ -8,53 +8,56 @@ import (
 	"github.com/mmcdole/gofeed"
 )
 
-// GetTagsFromRss gets all tags found in the RSS feed of a GitHub releases page
+// GetTagsFromRSS gets all tags found in the RSS feed of a GitHub releases page
 // Tags are sorted by descending creation date
-func (c Client) GetTagsFromRss(releasesRssUrl string) ([]string, error) {
-	feed, err := c.getReleaseRssFeed(releasesRssUrl)
+func (c Client) GetTagsFromRSS(releasesRSSURL string) ([]string, error) {
+	feed, err := c.getReleaseRSSFeed(releasesRSSURL)
 	if err != nil {
 		return nil, err
 	}
 
 	var tags = make([]string, 0)
 	for _, item := range feed.Items {
-		tag, err := c.extractTag(item)
-		if err != nil {
-			return nil, err
-		}
+		tag := c.extractTag(item)
 		if tag != nil {
 			tags = append(tags, *tag)
 		}
+
+		tags = append(tags, *tag)
 	}
 
 	return tags, nil
 }
 
-func (c Client) extractTag(item *gofeed.Item) (*string, error) {
-	pattern := regexp.MustCompile(`.*/(?P<Version>[a-zA-Z0-9.\-_+]+)$`)
-	matches := pattern.FindStringSubmatch(item.GUID)
+// tagPattern is used in extractTag to extract the tag from the RSS item
+var tagPattern = regexp.MustCompile(`.*/(?P<Version>[a-zA-Z0-9.\-_+]+)$`)
+
+func (c Client) extractTag(item *gofeed.Item) *string {
+	matches := tagPattern.FindStringSubmatch(item.GUID)
 
 	if matches == nil {
 		c.log.Warn(fmt.Sprintf("Could not parse RSS item %s", item.Link))
-		return nil, nil
+		return nil
 	}
 
-	return &matches[pattern.SubexpIndex("Version")], nil
+	return &matches[tagPattern.SubexpIndex("Version")]
 }
 
-func (c Client) getReleaseRssFeed(releasesRssUrl string) (*gofeed.Feed, error) {
-	done := c.rssThrottle()
-	defer done()
-
-	resp, err := c.httpClient.Get(releasesRssUrl)
+func (c Client) getReleaseRSSFeed(releasesRSSURL string) (*gofeed.Feed, error) {
+	resp, err := c.httpClient.Get(releasesRSSURL)
 	if err != nil {
-		return nil, fmt.Errorf("%s: %w", releasesRssUrl, err)
+		return nil, fmt.Errorf("%s: %w", releasesRSSURL, err)
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("%s got error %d", releasesRssUrl, resp.StatusCode)
+		return nil, fmt.Errorf("%s got error %d", releasesRSSURL, resp.StatusCode)
 	}
 
-	return gofeed.NewParser().Parse(resp.Body)
+	feed, err := gofeed.NewParser().Parse(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse rss feed, %s: %w", releasesRSSURL, err)
+	}
+
+	return feed, nil
 }
--- a/src/internal/github/rss.go
+++ b/src/internal/github/rss.go
@@ -44,6 +44,9 @@ func (c Client) extractTag(item *gofeed.Item) *string {
 }
 
 func (c Client) getReleaseRSSFeed(releasesRSSURL string) (*gofeed.Feed, error) {
+	done := c.rssThrottle()
+	defer done()
+
 	resp, err := c.httpClient.Get(releasesRSSURL)
 	if err != nil {
 		return nil, fmt.Errorf("%s: %w", releasesRSSURL, err)
--- a/src/internal/github/rss.go
+++ b/src/internal/github/rss.go
@@ -22,8 +22,6 @@ func (c Client) GetTagsFromRSS(releasesRSSURL string) ([]string, error) {
 		if tag != nil {
 			tags = append(tags, *tag)
 		}
-
-		tags = append(tags, *tag)
 	}
 
 	return tags, nil
--- a/src/internal/module/list.go
+++ b/src/internal/module/list.go
@@ -38,27 +38,24 @@ func extractModuleDetailsFromPath(path string) *Module {
 	return &m
 }
 
-type ModuleList []Module
-
-func ListModules(moduleDataDir string, logger *slog.Logger, ghClient github.Client) (ModuleList, error) {
-	// walk the module directory recursively and find all json files
-	// for each json file, parse it into a Module struct
-	// return a slice of Module structs
-
+// ListModules walks the module metadata directory provided and returns a list of modules.
+func ListModules(moduleDataDir string, logger *slog.Logger, ghClient github.Client) (List, error) {
 	var results []Module
 	err := filepath.Walk(moduleDataDir, func(path string, info os.FileInfo, err error) error {
 		m := extractModuleDetailsFromPath(path)
-		if m != nil {
-			m.Directory = moduleDataDir
-			m.Logger = logger.With(
-				slog.String("type", "module"),
-				slog.Group("module", slog.String("namespace", m.Namespace), slog.String("name", m.Name), slog.String("targetsystem", m.TargetSystem)),
-			)
-			m.Github = ghClient.WithLogger(m.Logger)
-			results = append(results, *m)
-		} else {
+		if m == nil {
 			logger.Debug("Failed to extract module details from path, skipping", slog.String("path", path))
+			return nil
 		}
+
+		// enrich the module with additional information
+		m.Directory = moduleDataDir
+		m.Logger = logger.With(
+			slog.String("type", "module"),
+			slog.Group("module", slog.String("namespace", m.Namespace), slog.String("name", m.Name), slog.String("targetsystem", m.TargetSystem)),
+		)
+		m.Github = ghClient.WithLogger(m.Logger)
+		results = append(results, *m)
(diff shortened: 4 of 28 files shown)


cube2222 (approved):

src/internal/github/client.go
```diff
@@ -11,14 +11,18 @@ import (
...
 )
 
+const UserAgent = "OpenTofu Registry/1.0"
+
+// EnvAuthToken returns the GitHub token from the environment.
 func EnvAuthToken() (string, error) {
 	token := os.Getenv("GH_TOKEN")
 	if token == "" {
-		return "", fmt.Errorf("Expected $GH_TOKEN to be set, unable to authenticate with GitHub")
+		return "", fmt.Errorf("expected $GH_TOKEN to be set, unable to authenticate with GitHub")
```
ghost: maybe `expected the GH_TOKEN environment variable to be set` to make it explicit?

src/internal/provider/provider.go
```diff
@@ -42,21 +46,23 @@ type Provider struct {
...
 
-// TODO custom repository url?
+// RepositoryURL returns the URL of the repository that the provider is assumed to be living in.
 func (p Provider) RepositoryURL() string {
 	return fmt.Sprintf("https://github.com/%s/%s", p.EffectiveNamespace(), p.RepositoryName())
 }
 
-func (p Provider) getRssUrl() string {
+// RSSUrl returns the URL of the RSS feed for the repository's releases.
+func (p Provider) RSSUrl() string {
```
ghost: If we are capitalizing RSS, why nor URL?

src/internal/provider/shasum.go
```diff
@@ -5,6 +5,8 @@ import (
 	"strings"
 )
 
+// GetShaSums will attempt to download the SHA checksums file from the given URL and return a
+// map of file names to SHA checksums.
 func (p Provider) GetShaSums(shaFileDownloadUrl string) (map[string]string, error) {
```
ghost: Missing capitalization here? While we are at it?

src/internal/provider/update.go
```diff
@@ -52,9 +54,11 @@ func (p Provider) shouldUpdateMetadataFile() (bool, error) {
 
 }
 
-func (p Provider) getRssSemverTags() ([]string, error) {
-	releasesRssUrl := p.getRssUrl()
-	tags, err := p.Github.GetTagsFromRss(releasesRssUrl)
+// getSemverTagsFromRSS returns a list of semver tags from the RSS feed
+// ignoring all non-valid semver tags
+func (p Provider) getSemverTagsFromRSS() ([]string, error) {
```
ghost: SemVer capitalization?

src/internal/provider/update.go
```diff
@@ -52,9 +54,11 @@ func (p Provider) shouldUpdateMetadataFile() (bool, error) {
...
+	releasesRssUrl := p.RSSUrl()
```
ghost: RSS capitalization? URL capitalization?

ghost (approved):

src/internal/files/files.go
```diff
@@ -7,9 +7,9 @@ import (
 	"path"
 )
 
-// SafeWriteObjectToJSONFile writes the given object to the given file path.
+// SafeWriteObjectToJSONFile writes the given data to the given file path.
 // It also ensures that the destination directory exists and that the file is written correctly.
-func SafeWriteObjectToJSONFile(filePath string, data interface{}) error {
+func SafeWriteObjectToJSONFile(filePath string, data any) error {
```
kislerdm, line 12: **(question)** I feels like we are settling on a particular style - shall we make it a rule: use `any` instead of `interface{}` for new codebase? WDYT @opentofu/core-engineers?

src/internal/github/client.go
```diff
@@ -11,14 +11,18 @@ import (
 	"github.com/shurcooL/githubv4"
 )
 
+const UserAgent = "OpenTofu Registry/1.0"
```
kislerdm, line 14: **(question)** What is the reason for using version `1.0` here? Also, could you please share a (preliminary) vision on how we'd go about registry versioning? Thanks!

src/internal/github/client.go
```diff
@@ -11,14 +11,18 @@ import (
...
 func EnvAuthToken() (string, error) {
 	token := os.Getenv("GH_TOKEN")
 	if token == "" {
-		return "", fmt.Errorf("Expected $GH_TOKEN to be set, unable to authenticate with GitHub")
+		return "", fmt.Errorf("expected $GH_TOKEN to be set, unable to authenticate with GitHub")
 	}
 	return token, nil
 }
 
+// Client is a GitHub client that abstracts away the different GitHub APIs and handles rate limiting/throttling.s
```
kislerdm: **(nitpick)** @Yantrio Hey James! Is the dot intentional `limiting/throttling.s`?

src/internal/github/rss.go
```diff
@@ -29,15 +29,17 @@ func (c Client) GetTagsFromRss(releasesRssUrl string) ([]string, error) {
...
 
 	if matches == nil {
 		return "", fmt.Errorf("could not parse RSS item %s", item.Link)
 	}
 
-	return matches[pattern.SubexpIndex("Version")], nil
+	return matches[tagPattern.SubexpIndex("Version")], nil
 }
 
 func (c Client) getReleaseRssFeed(releasesRssUrl string) (*gofeed.Feed, error) {
```
kislerdm: **(nitpick)** Although not part of the changes, I'd like to suggest the following for the sake of consistency:

```suggestion
func (c Client) getReleaseRSSFeed(releasesRssUrl string) (*gofeed.Feed, error) {
```

src/internal/module/module.go
```diff
@@ -11,23 +11,28 @@ import (
...
-	TargetSystem string // The module target system
-	Directory    string // The root directory that the module lives in
-	Logger       *slog.Logger
-	Github       github.Client
+	Namespace    string        // The module namespace
+	Name         string        // The module name
+	TargetSystem string        // The module target system
+	Directory    string        // The root directory that the module lives in
+	Logger       *slog.Logger  // A logger for the module
+	Github       github.Client // A GitHub client for the module
```
kislerdm, line 32: **(suggestion)** Sorry for picking, but I feel 8 out of 10 on this one 😄 

```suggestion
        // The module namespace
	Namespace    string       
        // The module name
	Name         string  
        // The module target system
	TargetSystem string   
	// The root directory that the module lives in     
	Directory    string
        // A logger for the module
	Logger       *slog.Logger  
        // A GitHub client for the module
	Github       github.Client
```

**(question/todo)** What is the benefit of having `Logger` and `Github` as module's attributes? IMO module as an entity should not be coupled to Github, or Logger. Maybe something for another refactoring iteration? 😃

src/internal/parallel/each.go
```diff
@@ -1,7 +1,12 @@
 package parallel
 
+// Action is a function that returns an error, this is used by ForEach
```
kislerdm: **(question)** I find the comment redundant because it does not clarify the objective of this interface beyond sharing its signature and linking it to the function `ForEach`. Those are obvious conclusions coming from the code itself which don't require additional comment IMO. WDYT?

```suggestion
// Action defines the interface to abstract a single operation executed concurrently.
```

src/internal/parallel/each.go
```diff
@@ -1,7 +1,12 @@
...
 type Action func() error
 
+// ForEach runs the given actions in parallel with a maximum concurrency of maxConcurrency
+// and returns a slice of errors
```
kislerdm:
```suggestion
// ForEach runs the given Actions concurrently with a specified maximum concurrency.
```

Motivation: no need to repeat the function signature IMO. WDYT?

src/internal/parallel/each_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestForEach(t *testing.T) {
+	// Note: these actions are not thread-safe, but that's fine for this test
+	// The idea of this test is to ensure that the concurrency is working as expected
+	// and that the errors are returned as expected
+
+	successAction := func() error {
+		time.Sleep(50 * time.Millisecond) // simulate some work
```
kislerdm: **(nitpick)** WDYT about setting sleep to 1 ms, 50 ms feels too much of a waiting given the objectives of this test?

src/internal/parallel/each_test.go
```diff
@@ -0,0 +1,58 @@
...
+			actions:        []Action{errorAction, successAction, errorAction},
+			maxConcurrency: 2,
+			expectedErrs:   2,
+		},
+		{
+			name:           "Some actions with errors, more actions than concurrency",
+			actions:        []Action{errorAction, successAction, errorAction, errorAction, errorAction, errorAction},
+			maxConcurrency: 2,
+			expectedErrs:   5,
+		},
```
kislerdm, line 49: **(question)** Don't these test cases evaluate the same scenario? I feel that maybe using errors as an indication that all actions were running is not the best approach, WDYT? It maybe worth an adjustment of the `ForEach` signature:

```go
type ForEach(actions []Action, maxConcurrency int) []Status

type Status struct {
    Action Action
    Error  error
}
```

src/internal/util.go
```diff
@@ -2,6 +2,8 @@ package internal
 
 import "strings"
 
+// TODO: Move this somewhere to it's own package, not in internal.
+
```
kislerdm, line 6: **(suggestion)** I'd propose to move the function to the `github`, or `version` package.

src/internal/v1api/responses.go
```diff
@@ -13,47 +13,50 @@ type ModuleVersionListingResponse struct {
...
 
 type ModuleProviderDependency struct {
-	Name      string `json:"name"`
-	Namespace string `json:"namespace"`
-	Version   string `json:"version"` // The version constraint defined inside the module, ie. ">= 1.0.0"
-	Source    string `json:"source"`  // The name of the provider, ie. "hashicorp/aws" or "myregistry.com/myorg/myprovider"
+	Name      string `json:"name"`      // The name of the module
+	Namespace string `json:"namespace"` // The namespace of the module
+	Version   string `json:"version"`   // The version constraint defined inside the module, ie. ">= 1.0.0"
+	Source    string `json:"source"`    // The name of the provider, ie. "hashicorp/aws" or "myregistry.com/myorg/myprovider"
```
kislerdm, line 40:
```suggestion
	// The name of the module
	Name      string `json:"name"`
	// The namespace of the module
	Namespace string `json:"namespace"`
        // The version constraint defined inside the module, ie. ">= 1.0.0"
	Version   string `json:"version"`
	// The name of the provider, ie. "hashicorp/aws" or "myregistry.com/myorg/myprovider"
	Source    string `json:"source"`
```

src/internal/v1api/responses.go
```diff
@@ -13,47 +13,50 @@ type ModuleVersionListingResponse struct {
...
 
 type ModuleMetadata struct {
 	Path         string                     `json:"path,omitempty"` // If this is a submodule, the path to the module root
-	Providers    []ModuleProviderDependency `json:"providers"`
-	Dependencies []ModuleDependency         `json:"dependencies"`
+	Providers    []ModuleProviderDependency `json:"providers"`      // The providers required by this module
+	Dependencies []ModuleDependency         `json:"dependencies"`   // The modules required by this module
 
-	SubModules []ModuleMetadata `json:"submodules"`
+	SubModules []ModuleMetadata `json:"submodules"` // The submodules of this module
```
kislerdm, line 33:
```suggestion
	// If this is a submodule, the path to the module root
	Path string `json:"path,omitempty"`
	// The providers required by this module
	Providers []ModuleProviderDependency `json:"providers"`
	// The modules required by this module
	Dependencies []ModuleDependency `json:"dependencies"`
	// The submodules of this module
	SubModules []ModuleMetadata `json:"submodules"`
```

src/internal/v1api/responses.go
```diff
@@ -13,47 +13,50 @@ type ModuleVersionListingResponse struct {
...
+	Source    string `json:"source"`    // The name of the provider, ie. "hashicorp/aws" or "myregistry.com/myorg/myprovider"
 }
 
 type ModuleDependency struct {
-	Name    string `json:"name"`
-	Source  string `json:"source"`
-	Version string `json:"version"`
+	Name    string `json:"name"`    // The name of the module
+	Source  string `json:"source"`  // The location of the module
+	Version string `json:"version"` // The version constraint defined inside the module, ie. ">= 1.0.0"
```
kislerdm, line 46:
```suggestion
        // The name of the module
	Name string `json:"name"`
	// The location of the module
	Source string `json:"source"`
	// The version constraint defined inside the module, ie. ">= 1.0.0"
	Version string `json:"version"`
```

src/internal/v1api/responses.go
```diff
@@ -13,47 +13,50 @@ type ModuleVersionListingResponse struct {
...
+	Versions []ProviderVersionResponseItem `json:"versions"` // A list of provider versions.
 }
 
 type ProviderVersionResponseItem struct {
-	Version   string            `json:"version"`   // The version number of the provider.
-	Protocols []string          `json:"protocols"` // The protocol versions the provider supports.
-	Platforms []github.Platform `json:"platforms"` // A list of platforms for which this provider version is available.
+	Version   string              `json:"version"`   // The version number of the provider.
+	Protocols []string            `json:"protocols"` // The protocol versions the provider supports.
+	Platforms []provider.Platform `json:"platforms"` // A list of platforms for which this provider version is available.
```
kislerdm:
```suggestion
        // The version number of the provider.
	Version string `json:"version"`
	// The protocol versions the provider supports.
	Protocols []string `json:"protocols"`
	// A list of platforms for which this provider version is available.
	Platforms []provider.Platform `json:"platforms"`
```

kislerdm (commented):
A great refactoring effort, kudos for comments! I left some feedback - would be happy to follow up on them. 😄

src/internal/module/module.go
cam72cam: Logger in this case as Namespace, Name, and TargetSystem automatically added to the log entry.  Github is in a similar boat, constructed from a common instance with a custom module or provider specific logger.

src/internal/v1api/responses.go
cam72cam: I prefer the original way personally, but could easily live with either.

src/internal/parallel/each_test.go
cam72cam: It might make more sense to make this a generic function, similar to @cube2222 's suggestion in the original PR that added this.  I didn't want to add that at the time, since it would have been the first generics added to the build/codebase.

cam72cam: The java equiv might look like: `public static <T> List<Pair<T, Error>> ForEach(List<T> items, Function<T, Error> action)`.  I need to spend some time learning about generics in go

cam72cam: It could also be done with any/casting as this is not currently used by a tight loop

src/internal/provider/platform.go
```diff
@@ -0,0 +1,7 @@
+package provider
+
+// Platform represents a platform that a provider supports.
```
cam72cam: Nitpick: Maybe this should live in v1api as it's only used in there?

cam72cam (approved):

src/internal/files/files.go
Yantrio: I actually forgot I did this. I'm happy either way. `any` is synonymous with `interface{}` anyway.

src/internal/github/client.go
Yantrio: I just used what @cam72cam had uploaded already and moved it here. I don't see any harm in 1.0 for now.

Also I feel like the registry doesn't really need versioning, so 1.0 will probably stay for a long time until we re-write?

src/internal/github/client.go
Yantrio: Nope!

src/internal/parallel/each.go
Yantrio: I've put some thought into this and I want to avoid the word `interface`

How do you feel about 

```go
// Action is the function to be consumed by ForEach. It should return an error if the action failed
// and should not be retried
```

src/internal/parallel/each.go
Yantrio: I removed the mentions of the function signature.

For now I've left the word `paralell` in there, but thats because it matches the package name. If we want to rename this later we can do in another PR.

src/internal/parallel/each_test.go
Yantrio: Sure 👍

src/internal/parallel/each_test.go
Yantrio: I also wrote these tests as a black-box style consumer of the method. The idea here is to catch regressions and define behavior.

@cam72cam can you raise an issue or link one here for re-working this, also please feel free to reach out and talk about generics. I'm happy to help.

src/internal/provider/platform.go
Yantrio: I agree, I'll move it

src/internal/provider/provider.go
Yantrio: we should, I'll do this now.

src/internal/files/files.go
Yantrio: I honestly don't mind either way.Can you open a discussion internally on slack to discuss and then we can go from there if this is something you feel strongly about.

src/internal/module/module.go
Yantrio: I really think that the comments being above the properties here is a very minor point and I would like to ensure we have flexibility on some of the placements. Especially for readability of files that just declare structs and nothing else.

Overall I think this is fine and I'm going to stand by my choice here.

src/internal/util.go
Yantrio: I agree, but let's move it later on. Honestly I think this can just be replaced with a direct call to strings.TrimPrefix. I'll put in another PR later to do that.

src/internal/github/client.go
cam72cam: That was from @RLRabinowitz's original work, and seemed like a reasonable value to let people know who to contact.

src/internal/github/rss.go
```diff
@@ -8,53 +8,59 @@ import (
...
 	return tags, nil
 }
 
-func (c Client) extractTag(item *gofeed.Item) (*string, error) {
-	pattern := regexp.MustCompile(`.*/(?P<Version>[a-zA-Z0-9.\-_+]+)$`)
-	matches := pattern.FindStringSubmatch(item.GUID)
+// tagPattern is used in extractTag to extract the tag from the RSS item
+var tagPattern = regexp.MustCompile(`.*/(?P<Version>[a-zA-Z0-9.\-_+]+)$`)
+
+func (c Client) extractTag(item *gofeed.Item) *string {
```
cam72cam, line 33: :+1:

src/internal/parallel/each_test.go
cam72cam: https://github.com/opentofu/registry-stable/issues/40

src/internal/module/list.go
```diff
@@ -87,7 +89,7 @@ func (modules ModuleList) Parallel(maxConcurrency int, action Action) error {
 
 	errs := parallel.ForEach(actions, maxConcurrency)
 	if len(errs) != 0 {
-		return fmt.Errorf("Encountered %d errors processing %d modules", len(errs), len(modules))
+		return fmt.Errorf("encountered %d errors processing %d l", len(errs), len(l))
```
cam72cam: Looks like a find/replace problem in this message.

Yantrio: Good catch, I'll fix this.

cam72cam (approved):

src/internal/parallel/each.go
kislerdm: Sounds good!

src/internal/files/files.go
kislerdm: > you feel strongly about

Not particularly :) I'd say let's go without particularly strong guidelines yet as it's a no-problem yet. WDYT?

src/internal/github/client.go
kislerdm: Thanks for clarification! Let's keep this thread for the docu purposes - I feel that it's not the last time when some of us would stumble upon this UA string and ask themself that question about the version 1.0 :)

kislerdm (approved):

PR #50, opened 2023-11-28
Use repository tags to skip git ls-remote

### LINKED ISSUE #49: Use tags.atom to prevent modules from re-generating when not needed
This is probably a lot faster than git ls-remote.  Additionally the output seems to imply each tag in there has a release, but that is false.  Example: https://github.com/opentofu/tags.atom

cam72cam:
I've been testing with a few 10s of thousands of modules and realized that with multiple runs, it's faster (orders of magnitude) to use the tags.atom feed the same way we do with providers.

Fixes #49


cam72cam:
This also makes me hate the copy/paste between modules/providers even more and makes me want to refactor it.  At this point though, that's probably code golfing.

cam72cam:
The only thing I haven't tested is how fast I can get the `git ls-remote` to work before github bans my IP address...

Yantrio (approved):

PR #51, opened 2023-11-28
added forgotten tags for platform props

Yantrio:
This fixes an issue where the json is serialized with uppercase characters.


cam72cam (approved):
I was wondering about that...

PR #52, opened 2023-11-28
Providers build should work even if the manifest file is broken

RLRabinowitz:
- Change the code not to fail if we've failed to parse the manifest JSON file - simply assume that the protocol is `5.0` in such a case
- Upload providers that failed the hydration process as part of https://github.com/opentofu/registry-stable/pull/46 due to manifest parsing errors
  - `coxedge` providers were manually set to 6.0, as that's the expected protocol there, even though the manifest files are broken (have trailing `,`)

### DIFF
--- a/src/internal/provider/protocols.go
+++ b/src/internal/provider/protocols.go
@@ -23,7 +23,7 @@ func (p Provider) GetProtocols(manifestDownloadUrl string) ([]string, error) {
 
 	manifest, err := parseManifestContents(contents)
 	if err != nil {
-		return nil, err
+		return nil, nil
 	}
 
 	return manifest.Metadata.ProtocolVersions, nil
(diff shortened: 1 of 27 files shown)


src/internal/provider/protocols.go
```diff
@@ -23,7 +23,7 @@ func (p Provider) GetProtocols(manifestDownloadUrl string) ([]string, error) {
 
 	manifest, err := parseManifestContents(contents)
 	if err != nil {
-		return nil, err
+		return nil, nil
```
cam72cam, line 28: Can we add the following log message?
```go
p.Logger.Warn("Ignoring invalid manifest file", slog.String("url", manifestDownloadUrl), slog.Any("err", err))
```

RLRabinowitz: a710b83069b3405d0f2d77b49f0b43849f7b7e72

Yantrio (approved):

PR #58, opened 2023-11-29
Ensure that providers use lowercase pathing in the v1 api

### LINKED ISSUE #57: Providers containing Upper Case characters don't work
If you try using `1Password/onepassword` using the registry, it would not work. OpenTofu lowercases the provider source when making the V1 API calls, so it calls `v1/providers/1password/onepassword/versions`. However, this file does not currently exist

However, modules **are** case-sensitive. `terraform-aws-modules/aws/iam` and `TERRAFORM-AWS-MODULES/aws/iam` would be considered different modules.

So, for providers only, we'd need to make sure that we host the files in lowercase folder names

Yantrio:
This is to ensure that we serve things up in the same way that the clients expect.

Fixes #57

### DIFF
--- a/src/internal/v1api/providers.go
+++ b/src/internal/v1api/providers.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log/slog"
 	"path/filepath"
+	"strings"
 
 	"github.com/opentofu/registry-stable/internal/files"
 	"github.com/opentofu/registry-stable/internal/provider"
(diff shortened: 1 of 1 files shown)


src/internal/v1api/providers.go
```diff
@@ -34,12 +35,14 @@ func NewProviderGenerator(p provider.Provider, destination string) (ProviderGene
 
 // VersionListingPath returns the path to the provider version listing file.
 func (p ProviderGenerator) VersionListingPath() string {
-	return filepath.Join(p.Destination, "v1", "providers", p.Provider.Namespace, p.Provider.ProviderName, "versions")
+	listingPath := filepath.Join(p.Destination, "v1", "providers", p.Provider.Namespace, p.Provider.ProviderName, "versions")
```
cam72cam: BUG: if the p.Destination has upper/lower cases it will break

Yantrio: Fixed, thanks

cam72cam (approved):

PR #59, opened 2023-11-29
Added functionality to read gpg keys from the keys directory

Yantrio:
This PR also adds the opentofu gpg keys as a sample to ensure this works.


src/internal/v1api/providers.go
```diff
@@ -69,6 +74,16 @@ func (p ProviderGenerator) VersionListing() ProviderVersionListingResponse {
...
 	versionDetails := make(map[string]ProviderVersionDetails)
 
+	keyCollection := gpg.KeyCollection{
+		Namespace: p.Provider.EffectiveNamespace(),
+		Directory: p.KeyLocation,
+	}
+
+	keys, err := keyCollection.ListKeys()
+	if err != nil {
+		p.log.Error("Failed to list keys", slog.Any("err", err))
```
cam72cam, line 84: Do we want to return the error?

cam72cam (approved):
Overall looks good to me, just not sure if we want to kill a provider if keys list does not work.

src/internal/v1api/providers.go
Yantrio: we probably should. I'll add this tomorrow morning. Good catch!

Yantrio:
I've added this, If we cannot read the gpg key for some reason, i feel like we shouldn't update the rendered v1 responses at all. I would rather have outdated than broken responses.

keys/o/opentofu/20230928.asc
```diff
@@ -0,0 +1,29 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
cube2222: I think we can get rid of this key?

Yantrio: Yup, we can. I'll remove it.

src/internal/gpg/keycollection.go
```diff
@@ -0,0 +1,44 @@
...
+	files, err := os.ReadDir(location)
+	if err != nil {
+		return nil, err
+	}
+
+	keys := make([]Key, len(files))
+	for i, file := range files {
+		key, err := buildKey(filepath.Join(location, file.Name()))
+		if err != nil {
+			return nil, err
```
cube2222: error wrap maybe for the given directory?

cube2222 (approved):

PR #60, opened 2023-11-29
Alpha registry backfill

cam72cam:
Looked at the stats from the alpha registry and pulled in modules + providers that were missed in the initial github search.

Also worked around some GoogleCloudPlatform -> terraform-google-modules issues.


cube2222 (approved):

cam72cam:
I'm going to have to double check that I didn't add any case sensitive duplicates before this is merged. [done]

Yantrio (approved):

PR #77, opened 2023-12-01
Log non-fatal accessing repos errors instead of failing 

### LINKED ISSUE #66: Be more tolerant of missing repositories when bumping versions

cam72cam:
Longer term we need to track these and likely remove them after a grace period

Fixes https://github.com/opentofu/registry/issues/66

I also reverted some previous changes to remove providers/modules that were having either temporary or permanent issues.


src/internal/module/update.go
```diff
@@ -53,11 +57,15 @@ func (p Module) getSemVerTagsFromRSS() ([]string, error) {
 func (p Module) getLastSemVerTag() (string, error) {
 	semverTags, err := p.getSemVerTagsFromRSS()
 	if err != nil {
-		return "", err
+		p.Logger.Error("Unable to fetch tags, skipping", slog.Any("err", err))
+		return "", nil
+		//return "", err
```
cube2222: leftover comment

src/internal/module/update.go
```diff
@@ -53,11 +57,15 @@ func (p Module) getSemVerTagsFromRSS() ([]string, error) {
...
 	}
 
 	if len(semverTags) < 1 {
-		return "", fmt.Errorf("no semver tags found in repository %s", p.RepositoryURL())
+		p.Logger.Error("no semver tags found in repository, skipping", slog.String("url", p.RepositoryURL()))
+		return "", nil
+		//return "", fmt.Errorf("no semver tags found in repository %s", p.RepositoryURL())
```
cube2222: leftover comment

src/internal/provider/update.go
```diff
@@ -77,11 +82,15 @@ func (p Provider) getSemVerTagsFromRSS() ([]string, error) {
 func (p Provider) getLastSemVerTag() (string, error) {
 	semverTags, err := p.getSemVerTagsFromRSS()
 	if err != nil {
-		return "", err
+		p.Logger.Error("Unable to fetch tags, skipping", slog.Any("err", err))
+		return "", nil
+		//return "", err
```
cube2222: leftover comment

src/internal/provider/update.go
```diff
@@ -77,11 +82,15 @@ func (p Provider) getSemVerTagsFromRSS() ([]string, error) {
...
 	}
 
 	if len(semverTags) < 1 {
-		return "", fmt.Errorf("no semver tags found in repository %s", p.RepositoryURL())
+		p.Logger.Error("no semver tags found in repository, skipping", slog.String("url", p.RepositoryURL()))
+		return "", nil
+		//return "", fmt.Errorf("no semver tags found in repository %s", p.RepositoryURL())
```
cube2222: leftover comment

src/internal/provider/update.go
```diff
@@ -77,11 +82,15 @@ func (p Provider) getSemVerTagsFromRSS() ([]string, error) {
 func (p Provider) getLastSemVerTag() (string, error) {
 	semverTags, err := p.getSemVerTagsFromRSS()
 	if err != nil {
-		return "", err
+		p.Logger.Error("Unable to fetch tags, skipping", slog.Any("err", err))
+		return "", nil
```
ghost, line 87: I would consider returning a "known error" here that can be checked with `errors.Is` instead of returning an empty tag with no error. From a Go convention standpoint I expect the function to return something meaningful if no error happened, which doesn't happen here. This seems like it should be handled on the caller side.

src/internal/module/update.go
cam72cam: I did this intentionally as I see it as a stopgap

cube2222: Could you add a TODO comment then?

cam72cam: Yep

src/internal/provider/update.go
cam72cam: I was keeping with the current "pattern" in the codebase that doesn't use custom error types.  I'm hoping to rewrite a lot of this code and was trying to keep the changeset minimal.

cam72cam: Not saying it's a good pattern...

ghost: Feel free to merge, I don't know the codebase enough to comment anything meaningful other than what I did.

cube2222 (approved):

PR #79, opened 2023-12-01
Provider/Module submission workflow

### LINKED ISSUE #34: Implement the flow to allow contributors to add their provider to the registry

### LINKED ISSUE #35: Implement the flow to allow contributors to add their module to the registry

cam72cam:
Fixes: https://github.com/opentofu/registry/issues/34
Fixes: https://github.com/opentofu/registry/issues/35
Via this RFC: https://github.com/opentofu/opentofu/issues/916

I worked off of @roni-frantchi's original branch that he used to prototype during the RFC process

I forked the registry and created some example issues / PRs: https://github.com/cam72cam/opentofu-registry/issues

When an issue is created or edited with the correct labels, it will kick off an action.  Depending on label, it will either hit the provider or module job.

Validation:
* Provider/module matches expected pattern
* Provider/module does not already exist in the repo
* Provider/module does not have a branch open already (duplicate issue)
* Provider/module has versions that can be detected via a targeted version-bump

If any validation fails, a message will be added to the issue.  The messages could definitely stand to be improved.

If all of the validation passes, a PR will be created with the bumped provider/module.  A comment will be added to the issue when this occurs.

There is a commented out section that might allow us to auto-assign it to the core team which I could not test on my fork.


.github/workflows/issue-to-pr.yaml
```diff
@@ -0,0 +1,189 @@
...
+        with:
+          go-version-file: './src/go.mod'
+      - name: Provider Issue to Pull Request
+        run: |
+          # Pull provider out of issue body
+          provider=$(echo "$BODY" | grep "### Repository" -A2 | tail -n1 | sed -e 's/[\r\n]//g')
+          echo "Detected Provider '$provider'"
+
+          # Check that provider matches expected pattern
+          if ! [[ "$provider" =~ [a-zA-Z0-9\-]+/terraform-provider-[a-zA-Z0-9\-]+ ]]; then
```
RLRabinowitz: This validation is probably good enough for now, but we might want to harden it further, to further align with the CLI.

For example, [it doesn't allow for multiple dashes in a row](https://github.com/opentofu/registry-address/blob/901b9ae4061a6e83e7db96e71b8d7088294930b3/provider.go#L422-L429)

.github/workflows/issue-to-pr.yaml
```diff
@@ -0,0 +1,189 @@
...
+        with:
+          go-version-file: './src/go.mod'
+      - name: Module Issue to Pull Request
+        run: |
+          # Pull module out of issue body
+          module=$(echo "$BODY" | grep "### Repository" -A2 | tail -n1 | sed -e 's/[\r\n]//g')
+          echo "Detected Module '$module'"
+
+          # Check that it's not a provider
+          if [[ "$module" =~ [a-zA-Z0-9\-]+/terraform-module-[a-zA-Z0-9\-]+ ]]; then
```
RLRabinowitz: This regex is incorrect. Did you mean for it to be as follows?
```suggestion
          if [[ "$module" =~ [a-zA-Z0-9\-]+/terraform-provider-[a-zA-Z0-9\-]+ ]]; then
```

.github/workflows/issue-to-pr.yaml
```diff
@@ -0,0 +1,189 @@
...
+            gh issue comment $NUMBER -b "Failed validation: Likely a provider '$module'"
+            exit 1
+          fi
+
+          # Check that module matches expected pattern
+          if ! [[ "$module" =~ [a-zA-Z0-9\-]+/terraform-[a-zA-Z0-9]+-[a-zA-Z0-9\-]+ ]]; then
```
RLRabinowitz: This validation is probably good enough for now, but we might want to harden it further, to further align with the CLI.

For example, the target part should be limited to 64 character. You can see the regexes that the CLI uses for validation [here](https://github.com/opentofu/registry-address/blob/901b9ae4061a6e83e7db96e71b8d7088294930b3/module.go#L39-L40)

RLRabinowitz (commented):
Overall LGTM

roni-frantchi:
Hey @cam72cam thanks for bringing this idea to life! ❤️  

I was hoping of getting to spin up a PR on top of this PR submitting some suggestions/carrying the torch for you till Monday but I think I might not get to it, so here are some thoughts:  

1. There's a lot of bash slicing and dicing getting the different form fields. I would use [this action for instance](https://github.com/marketplace/actions/github-issue-forms-body-parser) (there are others), which would get you the fields we want in a convenient `steps.parse.outputs.data.your-field-here`  
2. Maybe after (1)'s addressed it'll be less verbose, but I think extracting certain parts of the the job like validation etc to dedicated `sh` files would help view/review/test/maintain these as it'd be easier to run certain parts locally
3. Note that you've marked this flow for `edited` too, but on such occasion we might need to _edit_ an existing PR (or close and open another)
4. Need to add validation that DCO checkbox is checked (on edit too!). Could also add the DCO status check to the PR so that i'd be blocked for merge without it.  
5. Consider adding a `closes #{issue}` so that once the PR is merged it will automatically close the submission form issue
6. GPG keys - I think we settled on these being there on the provider form, which would open a PR for GPG submission (one that follows the current manual process, to not kick in a new process).  

Also wanted to say this vv
> section that might allow us to auto-assign it to the core team

is a **very** neat touch! I would consider doing so on failures only tho

roni-frantchi (commented):
Hey @cam72cam thanks for bringing this idea to life! ❤️  

I was hoping of getting to spin up a PR on top of this PR submitting some suggestions/carrying the torch for you till Monday but I think I might not get to it, so here are some thoughts:  

1. There's a lot of bash slicing and dicing getting the different form fields. I would use [this action for instance](https://github.com/marketplace/actions/github-issue-forms-body-parser) (there are others), which would get you the fields we want in a convenient `steps.parse.outputs.data.your-field-here`  
2. Maybe after (1)'s addressed it'll be less verbose, but I think extracting certain parts of the the job like validation etc to dedicated `sh` files would help view/review/test/maintain these as it'd be easier to run certain parts locally
3. Note that you've marked this flow for `edited` too, but on such occasion we might need to _edit_ an existing PR (or close and open another)
4. Need to add validation that DCO checkbox is checked (on edit too!). Could also add the DCO status check to the PR so that i'd be blocked for merge without it.  
5. Consider adding a `closes #{issue}` so that once the PR is merged it will automatically close the submission form issue
6. GPG keys - I think we settled on these being there on the provider form, which would open a PR for GPG submission (one that follows the current manual process, to not kick in a new process).  

Also wanted to say this vv
> section that might allow us to auto-assign it to the core team

is a **very** neat touch! I would consider doing so on failures only tho

roni-frantchi:
@cam72cam [this](https://github.com/marketplace/actions/issueops-validator) would also come in handy, looks like a lot of what we're trying to do here is handled in the advanced flow

roni-frantchi:
@cam72cam I've submitted a [stacked PR ](https://github.com/opentofu/registry/pull/80) capturing my thoughts and suggestions

.github/workflows/issue-to-pr.yaml
```diff
@@ -0,0 +1,189 @@
...
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: './src/go.mod'
+      - name: Module Issue to Pull Request
+        run: |
+          # Pull module out of issue body
```
ghost: This script is fairly long and may benefit from being externalized. That way we could test it separately with various inputs. Alternatively, this could be rewritten into something more robust than a shell script.

.github/workflows/issue-to-pr.yaml
ghost: Shouldn't this end on a `$`?

.github/workflows/issue-to-pr.yaml
```diff
@@ -0,0 +1,189 @@
...
+          grep -c '"version":' $jsonfile
+          if [[ "$?" != 0 ]]; then
+            gh issue comment $NUMBER -b "Failed validation: No versions detected in $jsonfile"
+            exit 1
+          fi
+          set -e
+
+          
+          git config --global user.email "no-reply@opentofu.org"
+          git config --global user.name "OpenTOFU Automation"
```
ghost:
```suggestion
          git config --global user.name "OpenTofu Automation"
```

.github/workflows/issue-to-pr.yaml
cube2222: I'm a big fan of writing things in go and just `go run`'ing them in GitHub Actions. We also have https://github.com/opentofu/scripts which does a lot of stuff with the github api (and uses Go instead of shell scripts).

RLRabinowitz:
> Note that you've marked this flow for edited too, but on such occasion we might need to edit an existing PR (or close and open another)

If the intention here was to support a case where a label was added to the issue, then we could probably use the `label` event instead. But I imagine that, for now, we'd only want to act on new issues

cam72cam:
> > Note that you've marked this flow for edited too, but on such occasion we might need to edit an existing PR (or close and open another)
> 
> If the intention here was to support a case where a label was added to the issue, then we could probably use the `label` event instead. But I imagine that, for now, we'd only want to act on new issues

It was intended to allow users to edit issues that fail validation instead of completely resubmitting.  One potential path is to lock the issue once the PR has been created.

.github/workflows/issue-to-pr.yaml
cam72cam: Pulled in that dependency.

.github/workflows/issue-to-pr.yaml
cam72cam: Pulled in that dependency.

RLRabinowitz:
> It was intended to allow users to edit issues that fail validation instead of completely resubmitting. One potential path is to lock the issue once the PR has been created.

Alternatively, we could just auto-close the issue if it fails validation. Not sure if a contributor would figure out that if they edit the issue it'll re-trigger the flow to submit the PR

IMO auto-close makes sense, would help us to avoid having garbage issues / PRs up that the core maintenance team would have to address

Yantrio:
Personally I feel like we should ask people to edit their uploads, I wouldn't like to keep filling out the same form over and over :(

Yantrio (approved):
I like the usage of registry-address package here for validation. 👍

src/cmd/add-module/main.go
```diff
@@ -0,0 +1,117 @@
...
+
+	jsonErr := files.SafeWriteObjectToJSONFile(*outputFile, output)
+	if jsonErr != nil {
+		// This really should not happen
+		panic(jsonErr)
+	}
+
+	if err != nil {
+		os.Exit(1)
+	}
```
Yantrio:
```suggestion
	if err != nil {
		logger.Error("Unable to add module", slog.Any("err", err))
		output.Validation = err.Error()
		os.Exit(1)
	}

	jsonErr := files.SafeWriteObjectToJSONFile(*outputFile, output)
	if jsonErr != nil {
		// This really should not happen
		panic(jsonErr)
	}
```

src/cmd/add-provider/main.go
```diff
@@ -0,0 +1,114 @@
...
+
+	jsonErr := files.SafeWriteObjectToJSONFile(*outputFile, output)
+	if jsonErr != nil {
+		// This really should not happen
+		panic(jsonErr)
+	}
+
+	if err != nil {
+		os.Exit(1)
+	}
```
Yantrio:
```suggestion
	if err != nil {
		logger.Error("Unable to add provider", slog.Any("err", err))
		output.Validation = err.Error()
		os.Exit(1)
	}

	jsonErr := files.SafeWriteObjectToJSONFile(*outputFile, output)
	if jsonErr != nil {
		// This really should not happen
		panic(jsonErr)
	}
```

src/cmd/add-module/main.go
cam72cam: We don't actually want to exit yet as we still need to write the json validation output

src/cmd/add-provider/main.go
cam72cam: We don't actually want to exit yet as we still need to write the json validation output

src/cmd/add-module/main.go
Yantrio: Okay 👍

cube2222 (approved):

PR #86, opened 2023-12-05
Dont remove objects from bucket on sync

### LINKED ISSUE #65: Disable deletion in default sync script, require workflow_dispatch with approval to remove objects from bucket.
So if we mess up our generation script, we don't wipe the bucket.

Use `--max-delete 0` in the default workflow.

Evi1Pumpkin:
By default, we don't want our R2 sync process to delete objects from the bucket. 
1. I added the flags `--max-delete 0 --delete-after` to the existing workflow `generate-and-deploy.yml`.
It's worth mentioning that if the `rclone sync` tries to delete objects from the bucket, the workflow will fail with an error. If it bugs us, we might consider switching to `rclone copy` instead.
2. In addition, I added a new workflow that allows the deletion of objects in case we want to overwrite the objects in the bucket. This workflow will use a new environment that will require 2 approvals.
3. Also changed the GH environments to represent production instead of development.

Fully tested it [on my fork](https://github.com/Evi1Pumpkin/registry).

Fixes https://github.com/opentofu/registry/issues/65

### DIFF
--- a/.github/workflows/generate-and-deploy-with-delete.yml
+++ b/.github/workflows/generate-and-deploy-with-delete.yml
@@ -1,0 +1,38 @@
+name: Generate and Sync Files with Delete
+
+on:
+  workflow_dispatch
+
+jobs:
+  generate-and-sync:
+    runs-on: ubuntu-latest
+
+    environment: Development-with-delete
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: './src/go.mod'
+
+      - name: Setup dependencies
+        run: sudo apt-get install rclone
+
+      - name: Run Generation script
+        working-directory: ./src
+        run: go run ./cmd/generate-v1 --destination ../generated
+        env:
+           GH_TOKEN: ${{ secrets.GH_TOKEN }}
+
+      - name: Sync Data to R2
+        run: rclone sync --checkers=512 --transfers=512 --checksum --fast-list ./generated R2:${{ secrets.R2_BUCKET_NAME }}
+        env:
+          # R2 credentials should be stored as GitHub secrets
+          RCLONE_CONFIG_R2_TYPE: s3
+          RCLONE_CONFIG_R2_PROVIDER: Cloudflare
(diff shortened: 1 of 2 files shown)


.github/workflows/generate-and-deploy-overwrite.yml
```diff
@@ -0,0 +1,38 @@
+name: Generate and Sync Files with Overwrite
+
+on:
+  workflow_dispatch
+
+jobs:
+  generate-and-sync:
+    runs-on: ubuntu-latest
+
+    environment: Development-with-overwrite
```
Evi1Pumpkin, line 10: We need to create this environment. It'll require 2 approvals.

.github/workflows/generate-and-deploy-with-delete.yml
```diff
@@ -0,0 +1,38 @@
+name: Generate and Sync Files with Delete
+
+on:
+  workflow_dispatch
+
+jobs:
+  generate-and-sync:
+    runs-on: ubuntu-latest
+
+    environment: Development-with-delete
```
Evi1Pumpkin: We'll need to create this environment with 2 approvals

cube2222: > Development-with-delete

I think this would indicate that this is a dev environment, right?

How about "Production-with-delete"? 😄

cube2222: I see the other env is also called this way, I'll add a "Production" env.

ronnyorot: Thanks, updated the the environments [here](https://github.com/opentofu/registry/pull/86/commits/925bbe4d4e1e9926b3ad9163afb375d032f5100a)

Evi1Pumpkin: Thanks. I updated the environments names [here](https://github.com/opentofu/registry/pull/86/commits/be393aae7327de651dff42badf6042cdfabfd98a).

Yantrio (approved):

cam72cam (approved):

cam72cam:
LGTM.  Only suggestion might be putting a warning at the top of both workflows to keep them in sync (other than the delete flag)

cube2222 (approved):
LGTM, nice!

PR #95, opened 2023-12-07
Create registry dev environment

### LINKED ISSUE #64: Implement a dev bucket and workflow to publish to the devregistry.

Evi1Pumpkin:
We want to create a dev environment for our registry, for easy development without impacting production.

### **The idea**
1. The dev environment will use a new DEV bucket.
2. We'll run the dev environment workflows through the Actions screen. The workflows will run separately, so we can test each part of the flow without triggering the whole process.
3. Before triggering a workflow we'll choose the workflow's branch and the environment from the dropdowns before each run:
Screenshot 2023-12-07 at 16 26 34

4. On dev environment, we'll commit and push changes in the registry's data to a new generated branch called `<original-branch-name>-bump-versions-<unix-timestamp>`. So we won't make the branch dirty with unwanted changes.

### **Changes**
1. Cron moved to `bump-and-publish-versions-prod.yml`.
2. `bump-versions.yml` and `generate-and-deploy.yml` can now be triggered as sub-workflows (by `bump-and-publish-versions-prod.yml`), or dispatched from the actions screen for development porpuses.
3. `bump-versions.yml` and `generate-and-deploy.yml` get the `environment` setting as an input.
4. If `bump-versions.yml` was triggered with the `development` env, it'll push the changes to a new branch.

### **How to use the registry dev environment**
I created a branch with changes and want to run it targeting the dev registry. The branch is called `Evi1Pumpkin-patch-1`.
1. I go to the Actions screen, choose the `Bump Provider and Module Versions` workflow, and run it using my new branch:
Screenshot 2023-12-11 at 21 53 37

2. The modules and providers' data is pushed to a new branch called `Evi1Pumpkin-patch-1-bump-versions-1702366742`. I can get it from the `Print Branch` step. If you have a better idea on how to make it more user-friendly I would love to hear.
Screenshot 2023-12-12 at 10 21 24
3. When I run `Generate and Sync Files` workflow, I supply the new branch as the source of the code and the providers and modules branch:
Screenshot 2023-12-12 at 17 47 44

Fixes https://github.com/opentofu/registry/issues/64


.github/workflows/bump-and-publish-versions-prod.yml
```diff
@@ -0,0 +1,24 @@
...
+        uses: benc-uk/workflow-dispatch@v1
+        with:
+          workflow: bump-versions.yml
+          inputs: '{"environment": "Production"}'
+
+      - name: Invoke publish workflow
+        uses: benc-uk/workflow-dispatch@v1
+        with:
+          workflow: generate-and-deploy.yml
+          inputs: '{"environment": "Production"}'
```
cube2222: I believe `benc-uk/workflow-dispatch@v1` invokes the workflow **asynchronously**. While we need generate-and-deploy.yml to happen **after** bump-versions.yml.

I'm not sure if there's an easy way around it, so it might be simpler to just keep the current workflows as-is, and create separate workflows for the workflow_dispatch actions.

Evi1Pumpkin: There is an easy way. If we drop `benc-uk/workflow-dispatch@v1` and instead use GH's [reused workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) we can wait for the first action to finish running (using the `needs` prop - as done [here](https://github.com/opentofu/registry/pull/95/files#diff-6a6c5216731ee30b6700ab88d1910736029a9f9342c097362a9a39b9fcac6072R15)).
I'm testing it on my fork and it's looking good.

.github/workflows/bump-versions.yml
```diff
@@ -28,12 +39,24 @@ jobs:
         working-directory: ./src
         run: go run ./cmd/bump-versions
 
+      - name: Get branch
+        id: get-branch
+        run: |
+          if [ "${{ inputs.environment }}" = "Production" ]; then
```
ghost, line 45: ~You are passing `inputs.environment` as the `name` environment variable above. You could use `${name}` instead of the GitHub expression here. I believe, it is generally preferred to pass in inputs as environment variables to avoid injection issues, but it doesn't matter in this case.~ I misread, I thought the environment variables are being passed in. I vaguely remember that there was a recommendation to pass in input variables as environment variables, but I can't find it anymore.

.github/workflows/bump-versions.yml
```diff
@@ -28,12 +39,24 @@ jobs:
...
+            export BRANCH="${{ github.ref }}-bump-versions-$TIMESTAMP"
+          fi
+          echo "BRANCH=$BRANCH" >> $GITHUB_ENV
+
       - uses: stefanzweifel/git-auto-commit-action@v5
         with:
           file_pattern: 'modules/**/* providers/**/*'
           commit_user_name: Auto Version Bumper
+          branch: ${{ env.BRANCH }}
+          create_branch: true
```
RLRabinowitz, line 58: Does this work correctly (not create a branch) when the branch already exists? Making sure this won't break the bump versions for prod

.github/workflows/generate-and-deploy-with-delete.yml
```diff
@@ -2,31 +2,60 @@
...
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.bump-versions-branch }}
+          path: providers-and-modules
 
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version-file: './src/go.mod'
+          go-version-file: './main/src/go.mod'
```
ghost: I got confused for a moment when I read this... maybe this could be renamed to something like `repo`?

.github/workflows/generate-and-deploy-with-delete.yml
```diff
@@ -2,31 +2,60 @@
...
+            DATA_DIR_FLAGS="--module-data $PATH_PREFIX/modules --provider-data $PATH_PREFIX/providers --key-data $PATH_PREFIX/keys"
+          fi
+          
+          go run ./cmd/generate-v1 --destination ../../generated $DATA_DIR_FLAGS
         env:
-           GH_TOKEN: ${{ secrets.GH_TOKEN }}
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
 
       - name: Sync Data to R2
         run: rclone sync --checkers=512 --transfers=512 --checksum --fast-list ./generated R2:${{ secrets.R2_BUCKET_NAME }}
```
ghost, line 39: Wouldn't this deploy to the same bucket regardless of which environment is selected?

.github/workflows/generate-and-deploy.yml
```diff
@@ -2,31 +2,65 @@
...
+            DATA_DIR_FLAGS="--module-data $PATH_PREFIX/modules --provider-data $PATH_PREFIX/providers --key-data $PATH_PREFIX/keys"
+          fi
+          
+          go run ./cmd/generate-v1 --destination ../../generated $DATA_DIR_FLAGS
         env:
-           GH_TOKEN: ${{ secrets.GH_TOKEN }}
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
 
       - name: Sync Data to R2
         run: rclone sync --checkers=512 --transfers=512 --checksum --fast-list --max-delete 0 --delete-after ./generated R2:${{ secrets.R2_BUCKET_NAME }}
```
ghost, line 44: Wouldn't this deploy to the same bucket regardless of which environment is selected?

.github/workflows/generate-and-deploy-with-delete.yml
RLRabinowitz: Was that done on purpose? Doesn't match the `go-version-file` in other GH workflows

.github/workflows/generate-and-deploy-with-delete.yml
```diff
@@ -2,31 +2,60 @@
...
-        run: go run ./cmd/generate-v1 --destination ../generated
+        working-directory: ./main/src
+        run: |
+          PATH_PREFIX="../../providers-and-modules"
+          
+          if [ -z "${{ inputs.bump-versions-branch }}" ]; then
+            DATA_DIR_FLAGS=""
+          else
+            DATA_DIR_FLAGS="--module-data $PATH_PREFIX/modules --provider-data $PATH_PREFIX/providers --key-data $PATH_PREFIX/keys"
+          fi
```
RLRabinowitz: Why is this conditional necessary? I don't see a reason those flags should be different if this is running on a custom branch or not

ghost (commented):
~I don't know the codebase enough to approve or request changes, but it seems to me that the upload would still happen to the same bucket. From what I read, this is not the intention? (Also, added some nitpicks.)~ Nevermind, I completely misread the code.

.github/workflows/bump-versions.yml
Evi1Pumpkin: Yep, I already tested it (here's an [example](https://github.com/Evi1Pumpkin/registry/actions/runs/7177197309/job/19543324980)).

.github/workflows/generate-and-deploy-with-delete.yml
Evi1Pumpkin: agree https://github.com/opentofu/registry/pull/95/commits/bd0e4cb3b9d3e0f2f0931cefe6e207d82b24104d

.github/workflows/bump-versions.yml
Evi1Pumpkin: Yes, the `environment` GH feature is quite confusing.

.github/workflows/generate-and-deploy-with-delete.yml
cam72cam: If I understand correctly, this is using the workflow branch for the ./repo directory and optionally allowing to override the providers/modules specified. via   I'm not sure when this is necessary?

.github/workflows/generate-and-deploy-with-delete.yml
Evi1Pumpkin: I had to change it because I changed the path of the checked-out code in [here](https://github.com/opentofu/registry/pull/95/files#diff-801c1b85939a1ea35b8b6e405143602ba9d9a08f04bf5280fa35ed59d2657f5eR33).
Before this change, we ran `git checkout` at root (=default active working directory) once:
root -> the code the WF will run (workflow + scripts) + the modules and providers.

Now the hierarchy is a little different:
root -> `repo` (I renamed `main`) -> the code the WF will run (workflow + scripts)
root -> `providers and modules` -> the branch where the modules and providers were pushed to.

That's why the code is inside the `repo` folder now, and we have to treat it accordingly.

.github/workflows/generate-and-deploy-with-delete.yml
Evi1Pumpkin: If we don't have a custom branch the source of our modules and providers is different. In this case, we'll use the providers and modules that exist on the WF's branch (`../providers` which is the default in the script if the flag wasn't provided).

.github/workflows/generate-and-deploy-with-delete.yml
RLRabinowitz: OK. Maybe a better approach would be to solely rely on the WF ref branch, and not get a custom branch for fetching the modules and providers.

This way, if you run the dev `bump-versions`, which creates a branch, then afterwards you can run the `generate-and-deploy` WF on the created branch. That branch will also include the PR's updated code

.github/workflows/generate-and-deploy-with-delete.yml
RLRabinowitz: Left [a suggestion](https://github.com/opentofu/registry/pull/95#discussion_r1423991407) for not having to use a custom branch or `./repo` directory

.github/workflows/generate-and-deploy-with-delete.yml
Evi1Pumpkin: True. Originally, I thought it'd be nicer if we had the code branch and the providers & modules branch separated. 
For example, I'm making a big change to both the `bump` script and the `generate and publish` script in my branch.
Let's say I run the `bump` script, have a new branch, and want to change something in the `generate and publish` script before I run it. If the branches are not separated, I'll have to remember to push my change to the newly created branch and my original branch, and it's something I can easily forget.
But I can see why it's too confusing. I'll go with your approach @RLRabinowitz

.github/workflows/generate-and-deploy-with-delete.yml
Evi1Pumpkin: I took your suggestion and reverted this part 🙃

.github/workflows/generate-and-deploy-with-delete.yml
Evi1Pumpkin: @RLRabinowitz implemented your suggetion

RLRabinowitz (approved):

PR #100, opened 2023-12-08
Fix links in README.md

rmartine-ias:
Links without a scheme are interpreted as relative file paths


cam72cam:
Good catch, thanks!

PR #108, opened 2023-12-14
More in depth refactoring of providers and modules

### LINKED ISSUE #30: Consider refactoring Module/Provider structs to split between identifier and data
See: https://github.com/opentofu/registry-stable/pull/26#discussion_r1402365176

cam72cam:
WIP

Fixes https://github.com/opentofu/registry/issues/30


src/internal/github/repository.go
```diff
@@ -0,0 +1,24 @@
+package github
+
+import (
+	"fmt"
+	"log/slog"
+)
+
+type Repository struct {
```
ghost, line 8: Since this seems to be called sometimes without `log`, maybe adding explicit constructor(s), even if not exported, would be beneficial? That would document what the required parameters are.

src/internal/github/client.go
```diff
@@ -46,30 +45,25 @@ func NewClient(ctx context.Context, log *slog.Logger, token string) Client {
...
-// WithLogger returns a new Client with the given logger.
-func (c Client) WithLogger(log *slog.Logger) Client {
-	return Client{
-		ctx:        c.ctx,
-		log:        log.WithGroup("github"),
-		httpClient: c.httpClient,
-		ghClient:   c.ghClient,
+func (c Client) Repository(owner, name string, log *slog.Logger) Repository {
+	return Repository{
+		client: Client{
```
ghost, line 55: Does this not need a `log` here? See my comment about constructors.

ghost (commented):
I wonder if adding a golangci-lint config and CI as part of this refactor would be a good thing. That way any issues with the code could now be caught early.

PR #123, opened 2023-12-20
Fix GitHub organization membership check, In case of personal GitHub organization

RLRabinowitz:
Check if username matches organization name, in case of personal GitHub organization, in the organization membership check


ghost (approved):
Ship it!

cube2222 (approved):

Issue #125, opened 2023-12-21
Provider: metal-stack-cloud/metal


vknabel:
### Provider Repository

metal-stack-cloud/terraform-provider-metal

### DCO
 
- [X] I sign this project's [DCO](https://developercertificate.org/)

cam72cam:
Thanks for the submission and the help correcting the process @vknabel !

PR #126, opened 2023-12-21
Fix submission namespace output when including dashes

vknabel:
When a new provider or module has been submitted which included a `-` within the namespace of the Git repository, only the last portion of the user or organization was picked.

See #125 as an example where `metal-stack-cloud/terraform-provider-metal` resulted in a errornous repository url of `https://github.com/cloud/terraform-provider-metal`.

### DIFF
--- a/src/cmd/add-module/main.go
+++ b/src/cmd/add-module/main.go
@@ -45,7 +45,7 @@ func main() {
 
 	err = func() error {
 		// Lower case input
-		re := regexp.MustCompile("(?P<Namespace>[a-zA-Z0-9]+)/terraform-(?P<Target>[a-zA-Z0-9]*)-(?P<Name>[a-zA-Z0-9-]*)")
+		re := regexp.MustCompile("(?P<Namespace>[a-zA-Z0-9-]+)/terraform-(?P<Target>[a-zA-Z0-9]*)-(?P<Name>[a-zA-Z0-9-]*)")
 		match := re.FindStringSubmatch(*repository)
 		if match == nil {
 			return fmt.Errorf("Invalid repository name: %s", *repository)
--- a/src/cmd/add-provider/main.go
+++ b/src/cmd/add-provider/main.go
@@ -44,7 +44,7 @@ func main() {
 
 	err = func() error {
 		// Lower case input
-		re := regexp.MustCompile("(?P<Namespace>[a-zA-Z0-9]+)/terraform-provider-(?P<Name>[a-zA-Z0-9-]*)")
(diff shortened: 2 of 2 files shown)


src/cmd/add-module/main.go
```diff
@@ -45,7 +45,7 @@ func main() {
 
 	err = func() error {
 		// Lower case input
-		re := regexp.MustCompile("(?P<Namespace>[a-zA-Z0-9]+)/terraform-(?P<Target>[a-zA-Z0-9]*)-(?P<Name>[a-zA-Z0-9-]*)")
+		re := regexp.MustCompile("(?P<Namespace>[a-zA-Z0-9-]+)/terraform-(?P<Target>[a-zA-Z0-9]*)-(?P<Name>[a-zA-Z0-9-]*)")
```
ghost, line 48: @cam72cam question: shouldn't this include a `/` prefix in the regexp itself and at least a `$` at the end? This regexp may match incomplete or incorrect repositories.

ghost:
@vknabel thank you for reporting this, it looks to me as if that regexp you changed could do with a little more attention from our side. It may need more than this change, but we'll try to sort it out as quickly as possible since this seems to be a blocker.

src/cmd/add-module/main.go
vknabel: Thanks for the fast feedback!

In case we use `^` and `$` to match the whole line, we also need to iterate over the comment line by line. (`^.+$` matches the whole multiline string)

e.g.

```go
re := regexp.MustCompile("^(?P<Namespace>[a-zA-Z0-9-]+)/terraform-(?P<Target>[a-zA-Z0-9]*)-(?P<Name>[a-zA-Z0-9-]*)$")
repoScanner := bufio.NewScanner(strings.NewReader(*repository))
var match []string
for len(match) == 0 && repoScanner.Scan() {
   match = re.FindStringSubmatch(repoScanner.Text())
}
```


Alternatives for matching the namespace taken from this repository:
* `[a-zA-Z0-9-]+` current _(too specific?)_
* `.*` in `internal/v1api/providers.go` `func ArchivedOverrides` _(too broad?)_
* `[^/]+` in `internal/provider/list.go` `providerDirectoryRegex` _(my favorite)_


Happy to implement your decision.

ghost: Thank you for all your help! I'm unfortunately not as well versed in the registry code as @cam72cam so I'll let him make the call on this. He should be online in a bit hopefully.

cam72cam: This should work fine.  The match below requires the full string to match.

PR #141, opened 2024-01-01
Improve rendering approach in the verification package.

### LINKED ISSUE #93: Improve rendering approach in the verification package.
We could improve this method as pointed out by @janosdebugs [here](https://github.com/opentofu/registry/pull/88#discussion_r1418650236)

We could use go templates, or lots of other different approaches.

prashant3286:
Fixes #93

### DIFF
--- a/src/pkg/verification/render.go
+++ b/src/pkg/verification/render.go
@@ -1,68 +1,49 @@
 package verification
 
-import (
-	"bytes"
-	"fmt"
-	"io"
-	"text/template"
-)
-
-var templates *template.Template
-
-func initTemplates() {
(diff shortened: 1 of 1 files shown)


Yantrio (changes requested):
Hi @prashant3286 

Thank you for the contribution, however it seems this this is non-functional

I'm not quite sure what is attempted here with regards to go templates.

This code also does not complile, doesn't pass unit tests, and doesn't even print anything as far as I can tell. 

Would you mind shedding some light on why you chose to do this in the way that you did?

cam72cam:
Closing as invalid / incomplete.

Issue #164, opened 2024-01-15
Add e2e tests on generated files.


cube2222:
This issue is two-fold.

The main point is that we should, after generating static registry files, spin up a local http server to mock the registry, and run some e2e tofu executions with it, to make sure it's working well.

This should be executed on generated files prior to syncing them to the prod bucket, and ideally also on each pull request.

kislerdm:
@cube2222 Hey Kuba, I'd be happy to give a hand with the issue this week.

cube2222:
Apologies for the late response, are you still interested in taking this @kislerdm?

Also, just to clarify the issue a bit more, the end to end tests should be very limited. One or two executions using some modules like local file. The goal is for this to not slow down the sync workflow too much.

The http server would ideally also be something simple off-the-shelf, like nginx serving on localhost. But if it’s not easy to do, it could also be a simple go binary.

virtualroot:
Do these jobs need to be fixed? Both workflows have zero runs:
- https://github.com/opentofu/registry/actions/workflows/validate-json-module.yml
- https://github.com/opentofu/registry/actions/workflows/validate-json-provider.yml

PR #174, opened 2024-01-17
Add validation for generated json metafiles

### LINKED ISSUE #152: Check for valid JSON on PRs

On auto-generated PRs we still should check  for valid JSON in case manual edits were necessary, see #147

kislerdm:
Resolves #152 

## What changed

- Added the program which validates json metadata files for modules and providers by performing a set of checks. It reports all found errors to stderr and terminates with the exit code 1 if the file is invalid, otherwise, it terminates with the exit code 0 without any messages.

## Why do we need it

- To validate correctness of updated json metafiles

#### Provider

image

#### Module

image


kislerdm:
Hey @janosdebugs @Yantrio @cam72cam! I'd appreciate your quick look over the change to see I'm going in the right direction with the implementation. Thanks!

Yantrio:
I think the implemtnation here is pretty solid, but I think it may be best to have an external type of validation here instead of using the same code that we're generating with.

Maybe it could be as simple as piping it into JQ and ensuring it's valid json?

@janosdebugs you raised the initial issue, when you mention "valid" json, do you mean that it's valid json, or valid to our format and what we expected?

ghost:
@Yantrio as I wrote in the related issue, I would like the JSON to be valid from the perspective of the registry, ideally. However, this may also be covered by #164.

kislerdm:
@janosdebugs @Yantrio Hey folks! Thanks for the feedback. On its basis, I'll remove introduced code and setup a github action which will read modified metafile using `jq`, and implement the content validation as part of #164.

ghost:
To clarify: I'm not unhappy with a Go solution, it feels more robust than shell scripting around it. That being said, I don't know the registry very well, so I'll have to rely on the more experienced colleagues for judgement here.

kislerdm:
@janosdebugs @Yantrio hey guys! The validator is ready for review.

.github/workflows/validate-json-module.yml
```diff
@@ -0,0 +1,45 @@
...
+          go-version-file: './src/go.mod'
+
+      - name: Build validator
+        working-directory: ./src
+        run: |
+          mkdir -p /tmp/validate
+          go build -o /tmp/validate/run ./cmd/validate/main.go
+          chmod +x /tmp/validate/run
+
+      - name: List updated filed
```
Yantrio:
```suggestion
      - name: List updated files
```

.github/workflows/validate-json-provider.yml
```diff
@@ -0,0 +1,45 @@
...
+          go-version-file: './src/go.mod'
+
+      - name: Build validator
+        working-directory: ./src
+        run: |
+          mkdir -p /tmp/validate
+          go build -o /tmp/validate/run ./cmd/validate/main.go
+          chmod +x /tmp/validate/run
+
+      - name: List updated filed
```
Yantrio:
```suggestion
      - name: List updated files
```

Yantrio (approved):
thanks for this @kislerdm , I've added 2 tiny tiny nitpicks but overall im happy to approve.

kislerdm:
@Yantrio good catch, thank you James! 😄

kislerdm:
@janosdebugs Hey Janos! Could you please review the changes? Thanks!

ghost (approved):
Very nice and thorough work, thank you!

PR #272, opened 2024-03-02
feat: add concurrency to workflow

### LINKED ISSUE #163: "Bump Versions and Publish - Prod" should use a concurrency group.
Otherwise two workflows could overlap.

virtualroot:
Allow only one workflow at that time.

If a second workflow is triggered, it will be put in pending state, and only start once the first workflow finish.
If the workflow is triggered a third time while the first one is running, the second workflow is canceled, and the third takes over the pending state.

Fix #163.

### DIFF
--- a/.github/workflows/bump-and-publish-versions-prod.yml
+++ b/.github/workflows/bump-and-publish-versions-prod.yml
@@ -5,8 +5,11 @@ on:
     - cron: '5 * * * *'
   workflow_dispatch:
 
+# Use global lock
+#
+# https://github.com/opentofu/registry/pull/272#discussion_r1511031778
 concurrency:
-  group: ${{ github.head_ref }}
+  group: main
 
 jobs:
   bump-versions:
--- a/.github/workflows/bump-and-publish-versions-prod.yml
+++ b/.github/workflows/bump-and-publish-versions-prod.yml
@@ -5,6 +5,9 @@ on:
     - cron: '5 * * * *'
   workflow_dispatch:
 
+concurrency:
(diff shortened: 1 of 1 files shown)


Yantrio (approved):

.github/workflows/bump-and-publish-versions-prod.yml
```diff
@@ -5,6 +5,9 @@ on:
     - cron: '5 * * * *'
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.head_ref }}
```
cam72cam: Bump versions can put new commits on main, which subsequent invocations would run under. Do we want this to use a static string instead?

virtualroot: @cam72cam 

I chose the dynamic value since I noticed the `workflow_dispatch`.
If there are no plans to do runs from branches other than the main, I can hardcode.

cam72cam: I just took a closer look athttps://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value.

We may want to run it on branches as validation at some point, but for now a global lock should suffice.  We should probably add a brief comment about this conversation in the file.

virtualroot: I changed the value to `main`, added a comment in the commit message, and linked to this discussion.

cam72cam (approved):

virtualroot:
@cube2222 #163 was opened by you. Please have a look and merge if it's okay.

PR #276, opened 2024-03-05
Copy security.md from main opentofu repo

Yantrio:
This PR brings across a copy of the SECURITY.md file from the main https://github.com/opentofu/opentofu repo. This helps us in the case that someone has a security concern with the registry

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,62 @@
+# Security Reporting Process
+
+Please report any security issue via
+[Private Vulnerability Reporting](https://github.com/opentofu/opentofu/security/advisories/new) where the issue will be triaged appropriately.
+Thank you in advance for helping to keep OpenTofu secure.
+
+# Security Release Process
+
+OpenTofu is a large growing community of volunteers, users, and vendors. The OpenTofu community has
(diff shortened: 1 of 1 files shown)


ghost:
Hey @Yantrio this may be better placed in the `.github` repository so it (AFAIK) applies to other repos too.

Yantrio:
The issue with a .github repository is that we probably dont want to be attaching this security guidelines to the forked repos we have. We need to be explicit on where we add this

SECURITY.md
```diff
@@ -0,0 +1,62 @@
...
+The initial Product Security Team will consist of members of Steering Committee and Core Development Team. In the future we may
+decide to have a subset of maintainers work on security response given that this process is time
+consuming.
+
+## Disclosures
+
+### Private Disclosure Processes
+
+The OpenTofu community asks that all suspected vulnerabilities be privately and responsibly disclosed
+via the [reporting policy](README.md#reporting-security-vulnerabilities).
```
virtualroot, line 31:
```suggestion
via the [reporting policy](https://github.com/opentofu/opentofu/blob/main/README.md#reporting-security-vulnerabilities).
```

ghost (approved):

PR #292, opened 2024-03-08
Update datadog.json

jogarao96:
using this commit trying to fix datadog repo

Resulting an error unable to find provider 

│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider
│ hashicorp/terraform-provider-datadog: provider registry
│ registry.opentofu.org does not have a provider named
│ registry.opentofu.org/hashicorp/terraform-provider-datadog
│

### DIFF
--- a/providers/d/DataDog/datadog.json
+++ b/providers/d/DataDog/datadog.json
@@ -2489,105 +2489,105 @@
       "protocols": [
         "5.0"
       ],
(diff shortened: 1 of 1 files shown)


ghost:
Hi @jogarao96, this seems to be a problem in your OpenTofu configuration rather than the registry. You need to reference this provider as `DataDog/datadog` instead of `datadog` or `terraform-provider-datadog`. The `terraform-provider-` prefix is required by Terraform (which OpenTofu is based on), but gets removed by the registry in naming.

I'll be closing this pull request as it would break the registry, but feel free to open an issue or ask on Slack if you need further help.

Issue #386, opened 2024-04-11
Provider Key: ovh


rbeuque74:
### Provider Namespace

ovh

### Public Membership

- [X] I have made my membership public

### Provider GPG Key

```
-----BEGIN PGP PUBLIC KEY BLOCK-----

mQINBGNzuO8BEADCHgVYsQCRYG79K0qGyA4yDNCCUURlvwXnuzvU7oKyMDnqQquk
33vUXwEEC0lsOgmA/m6BEkBfmjwDSFa+QcStwulmh//UgU0yKFq3jP+NiONQdLxz
+h7yhSYUutLP2TegqF33+QTjf3OvXQI31ZQ5A76l3x9cfXtXUJmtvYP9Zyc0/++n
Xu/ibTT8ihAEZ8WFVLRUzCqM1T+xo0hkeTELm1m+G7A4d3ZdcehIZJCnpMXfpJHm
4nmfNSoWwrzB1abf0r7t4k5u897kZ1gG3djGUcVhzz+TUC7DhPJrxyfu9rkxN3PZ
EUMl3MS4bENZX7eIPtxVcA0dNGXha8FgHZBZOz6UF5wyaiLN4XZNIaYtUrG7vQvK
wWOTKqsP2gvvqqO7v8ckFFqyJwCMmkkjhtw3HPUsFCcMS4PVwalvNIxJi2Yi3/yU
DGbyRuP5hwC7qzORZ1hkSXFDJ4hY+5yk15I0MgH1FCrJoRsOKdUNwTiH5aN+gbik
axNk9Nn09+wgMQwXOxAnp8Q81POr5TVjdFuyocgb3rECgkurBQW15MNfcolNLOdC
PhIWVdzAGn4sy1KFj3F9K5KiN3UoIi6Ngle3H2fukbHPTUX7IPRtS9QaIxJ4BcV1
mHuTt9CH1O7L3Fnd9/JETKO5FGoz1+Eq1tt12jiZ3wG16bjgxSqZBo8f4QARAQAB
tCZvdmgtY2RzIChvdmgtY2RzKSA8b3BlbnNvdXJjZUBvdmgubmV0PokCVAQTAQgA
PhYhBCxERbuOyyPz5+N4M5v5EIHbdWYZBQJjc7jvAhsDBQkSzAMABQsJCAcCBhUK
CQgLAgQWAgMBAh4BAheAAAoJEJv5EIHbdWYZAH8P/23UwNItcUArEMh76jRPzzBM
[...]

PR #388, opened 2024-04-11
Add second provider key ovh/

### LINKED ISSUE #385: OVH key not valid for all providers in namespace

The key added for ovh in #346 / #347 is only valid for the provider ovh/ovh, but not for the other providers in the ovh namespace (ovh/eyaml, ovh/mimirtool, ovh/mixtool), leading to them beeing uninstallable:

```
Error while installing ovh/mimirtool v0.1.2: authentication signature from unknown issuer
```

CC @rbeuque74

### OpenTofu Version

```shell
OpenTofu v1.6.2
on linux_amd64
```

```hcl
terraform {
  required_providers {
    mimirtool = {
      source  = "ovh/mimirtool"
      version = "0.1.2"
    }
  }
}
```

rbeuque74:
Fixes #385 #386 #387


cam72cam:
We are double checking that our api generation code will handle this correctly and merge shortly.

PR #506, opened 2024-05-13
Use lowercase paths for all generated files to allow for case-insensitive url rewrite rules to be put in place

### LINKED ISSUE #503: owner field should not be case sensitive when submitting a module

When publishing a module using `{owner}/terraform-{target}-{name} the owner field should not be case sensitive as in the public hashicorp registry this is also not the case.

### OpenTofu Version

```shell
OpenTofu v1.7.1
on darwin_arm64
+ provider registry.opentofu.org/hashicorp/azurerm v3.103.1
+ provider registry.opentofu.org/hashicorp/random v3.5.1
```

Both should work

```hcl
module "storage" {
  source  = "cloudnationhq/sa/azure"
  version = "~> 0.1"
```

```hcl
module "storage" {
  source  = "CloudNationHQ/sa/azure"
  version = "~> 0.1"
```

Yantrio:
Resolves #503 

Once this is merged, it will regenerate files with lowercase names. We will introduce a case-insensitive url re-write rule on the side of cloudfront which will help match to the new lowercase file paths.

### DIFF
--- a/src/internal/v1api/providers.go
+++ b/src/internal/v1api/providers.go
@@ -46,14 +46,14 @@ func NewProviderGenerator(p provider.Provider, destination string, gpgKeyLocatio
 func (p ProviderGenerator) VersionListingPath() string {
 	namespacePath := strings.ToLower(p.Provider.Namespace)
 	providerNamePath := strings.ToLower(p.Provider.ProviderName)
-	return filepath.Join(p.Destination, "v1", "providers", namespacePath, providerNamePath, "versions")
+	return strings.ToLower(filepath.Join(p.Destination, "v1", "providers", namespacePath, providerNamePath, "versions"))
 }
 
 // VersionDownloadPath returns the path to the provider version download file.
 func (p ProviderGenerator) VersionDownloadPath(ver provider.Version, details ProviderVersionDetails) string {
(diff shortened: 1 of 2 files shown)


src/internal/v1api/providers.go
```diff
@@ -46,14 +46,14 @@ func NewProviderGenerator(p provider.Provider, destination string, gpgKeyLocatio
 func (p ProviderGenerator) VersionListingPath() string {
 	namespacePath := strings.ToLower(p.Provider.Namespace)
 	providerNamePath := strings.ToLower(p.Provider.ProviderName)
-	return filepath.Join(p.Destination, "v1", "providers", namespacePath, providerNamePath, "versions")
+	return strings.ToLower(filepath.Join(p.Destination, "v1", "providers", namespacePath, providerNamePath, "versions"))
```
cam72cam: This is not necessary dye to the above ToLower.  The module changes should probably follow a similar pattern.

cam72cam (changes requested):

src/internal/v1api/providers.go
Yantrio: ahh good point, I was lazy and didn't read it all :D let me do it this way

cam72cam (approved):

PR #508, opened 2024-05-13
Added comment to explain erroring workflow

### LINKED ISSUE #507: Nightly syncs are failing

- 

### OpenTofu Version

```shell
https://github.com/opentofu/registry/actions/workflows/generate-and-deploy.yml

These are failing with 

2024/05/13 00:23:49 ERROR : Got fatal error on delete: --max-delete threshold reached
2024/05/13 00:23:49 ERROR : Got fatal error on delete: --max-delete threshold reached
```

we should set a high max delete threshold to work around this.
```


-

### PR DESCRIPTION
**Pull request by Yantrio**
Closes #507

### DIFF
--- a/.github/workflows/generate-and-deploy.yml
+++ b/.github/workflows/generate-and-deploy.yml
@@ -42,6 +42,7 @@ jobs:
       ## When the max-delete threshold is reached at 0. This is an indication that the
       ## bucket is not in sync with the local files. This is expected behavior and is a signal that we should
       ## manually run the generate-and-deploy-with-delete.yml workflow to ensure the bucket is in sync.
+      ## The sync will have happened and new files will be updated in the bucket, but old files will not have been deleted.
       - name: Sync Data to R2
         run: rclone sync --checkers=512 --transfers=512 --checksum --fast-list --max-delete 0 --delete-after ./generated R2:${{ secrets.R2_BUCKET_NAME }}
         env:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: .github/workflows/generate-and-deploy.yml [hunk 1]
```diff
@@ -38,6 +38,10 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GH_TOKEN }}
 
+      ## Note for developers consuming this: This action is sometimes expected to error
+      ## When the max-delete threshold is reached at 0. This is an indication that the
+      ## bucket is not in sync with the local files. This is expected behavior and is a signal that we should
+      ## manually run the generate-and-deploy-with-delete.yml workflow to ensure the bucket is in sync.
```
cube2222, line 44: I would add that the sync was still supposed to happen before the error was raised, so new files and overrides should have succeeded.

cube2222 (approved):
with nit

.github/workflows/generate-and-deploy.yml
Yantrio: 👍

Yantrio: added in https://github.com/opentofu/registry/pull/508/commits/a349ca528b96375cc470d858c1ca3a33ebee4ebe

ghost (approved):

PR #531, opened 2024-05-16
Added Landing Zone modules from Azure and azurenoops

DenisBalan:
While using

https://github.com/azurenoops/terraform-azurerm-overlays-management-spoke
https://github.com/azurenoops/terraform-azurerm-overlays-management-hub

Encountered some missing imports from terraform registry into tofu one.

This PR imports several of them.

Please squash, thx.


ghost:
@Yantrio how did we miss this?

Yantrio:
Good question, I suspect we missed this because it does not match the naming pattern of a normal module registry as determined by hashicorp here: https://developer.hashicorp.com/terraform/language/style#module-repository-names

I suspect that the multiple dashes in the repo names (because they're overlay) is what caused us to miss these in our initial hydration.


I know we're adding 5 items here but I would prefer if we went through the form as that will help validate that each of these are compatible with the registry

@DenisBalan I know this may be frustrating but would you mind submitting this form for each of the items please? https://github.com/opentofu/registry/issues/new?assignees=&labels=module%2Csubmission&projects=&template=module.yml&title=Module%3A+

Thanks!

DenisBalan:
@Yantrio done.

fixes: #539 #537 #535 #533 #532

DenisBalan:
@Yantrio @janosdebugs got it, seems that there is automation in place and this PR is not needed.

generally speaking, current approach of managing modules via present repository - is a bit of a mess.

its very volatile, as new modules/versions will keep changing daily or even more often.

is there any roadmap or grand vision on how whole module & registry governance will be in future?

Yantrio:
>its very volatile, as new modules/versions will keep changing daily or even more often.

@DenisBalan once a module is in our registry, we update versions as they come in. It's only the initial addition of the module or provider to the registry that is manual. The rest is automated from here. Our registry scrapes github often to check for new versions and publishes them as they are discovered.

As for the content of this PR, I've merged the other PRs so I will close this one. If you have any other questions please feel free to reach out by opening another issue. Thanks!

PR #624, opened 2024-05-28
Make sure to sync providers with modified keys

### LINKED ISSUE #621: Regular registry update "pruning" logic does not take new GPG keys into account

When somebody uploads a GPG key, we are not syncing it in the next run, because we're only running against pruned providers/modules. This logic [here](https://github.com/opentofu/registry/blob/main/.github/workflows/generate-and-deploy-recent.yml#L38-L45) does not take new GPG keys into account. 

### OpenTofu Version

```shell
-
```

-

cam72cam:
Fixes https://github.com/opentofu/registry/issues/621 by explicitly marking all providers impacted by new/modified keys as needing to be synchronized.

This will require all providers to be lower case (should have been done a while ago):
```
for i in {2..5}; do
        find ./modules/ -maxdepth "$i" -mindepth "$i" -type d | grep "[A-Z]" | xargs -I foo echo "git mv foo \$(echo foo | tr '[:upper:]' '[:lower:]')" | bash
        find ./providers/ -maxdepth "$i" -mindepth "$i" -type d | grep "[A-Z]" | xargs -I foo echo "git mv foo \$(echo foo | tr '[:upper:]' '[:lower:]')" | bash
done
```


.github/workflows/generate-and-deploy-recent.yml
```diff
@@ -40,9 +40,15 @@ jobs:
           # Copy files changed into pruned directory
           git fetch origin
           git checkout origin/main
-          mkdir -p ./pruned/providers ./pruned/modules
-          git diff --name-only `git rev-list -1 --before="$(date -d '-1 hours' '+%Y-%m-%d %H:%M:%S')" HEAD` | xargs -I foo echo "mkdir -p ./pruned/\$(dirname foo); cp foo ./pruned/foo" | bash
-          echo "Swapping dirs"
+
+          rm -rf ./pruned/
```
Yantrio: not needed

Yantrio (approved):

PR #961, opened 2024-09-13
ci: update `actions/setup-go` to v5

wyardley:
- Update `actions/setup-go` to v5 (removes deprecation warnings about node 16 runtime)
- Remove `cat file | jq` in favor of `jq < file` in workflow

While it falls forward to 20 anyway, it generates a warning like this:
image

### DIFF
--- a/.github/workflows/bump-versions.yml
+++ b/.github/workflows/bump-versions.yml
@@ -38,7 +38,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
         with:
           go-version-file: './src/go.mod'
           cache-dependency-path: './src/go.sum'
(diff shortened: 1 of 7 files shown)


ghost (approved):

.github/workflows/bump-versions.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
```
wyardley, line 41: note: I can pin these to a hash as the main opentofu/opentofu repo has it if preferred.

cam72cam (approved):

.github/workflows/bump-versions.yml
cam72cam: We need an org-wide policy soon.  For now, let's keep the changes to this repo minimal.

PR #1012, opened 2024-09-24
Revert "Create provider key microsoft/"

ollevche:
This reverts commit 9bbe57f54d5ec046d7aea4dda3565ceecf235260.


ollevche:
Currently, we support key per namespace and not per provider. This PR reverts the microsoft/fabric key to make other microsoft providers usable. Fixes https://github.com/opentofu/opentofu/issues/2014

Yantrio (approved):

PR #1033, opened 2024-09-26
WIP: Idempotent module tags

ghost:


src/internal/module/module.go
```diff
@@ -15,6 +15,7 @@ import (
 // Version represents a single version of a module.
 type Version struct {
 	Version string `json:"version"` // The version number of the provider. Correlates to a tag in the module repository
+	Ref     string `json:"ref"`     // Identifier that pins the version to a particular point in history
```
ghost, line 18: This needs to be retrofitted into libregistry too.

Issue #836, opened 2024-08-30
add module "identiops/k3s/hcloud"


mogol07:
### Thank you for reporting an issue.

this module helps to deploy k3s cluster to Hetzner cloud provider

### OpenTofu Version

```shell
OpenTofu v1.8.1
on windows_amd64
```

### OpenTofu Configuration Files

```hcl
// Snippet of HCL that can be used to help reproduce this problem
// full tf file could fetched from github.com/identiops/terraform-hcloud-k3s/raw/main/examples/1Region_3ControlPlane_3Worker_Nodes/main.tf
module "cluster" {
  # source       = "github.com/identiops/terraform-hcloud-k3s?ref=3.0.6"
  source                 = "identiops/k3s/hcloud"
  version                = "3.0.6"
  hcloud_token           = var.hcloud_token           # INFO: Set via `export TF_VAR_hcloud_token=xyz`
  hcloud_token_read_only = var.hcloud_token_read_only # INFO: Set via `export TF_VAR_hcloud_token_read_only=abc`

  # Cluster Settings
  # ----------------
  delete_protection = true # Must be set to false + `terraform apply` before destroying the cluster via `terraform destory`!
  cluster_name      = "prod"
  default_location  = "nbg1"         # See available locations https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server#location
  default_image     = "ubuntu-22.04" # See `HCLOUD_TOKEN=XXXX; curl -H \"Authorization: Bearer $HCLOUD_TOKEN\" https://api.hetzner.cloud/v1/images | jq -r .images[].name | sort`
  k3s_version       = "v1.28.5+k3s1" # See available versions, https://update.k3s.io/v1-release/channels regular images: https://hub.docker.com/r/rancher/k3s/tags upgrade images: https://hub.docker.com/r/rancher/k3s-upgrade/tags

  # General Settings
  # ----------------
  ssh_keys = {
    "john" = file("~/.ssh/id_ed25519.pub")
    "jane" = "ssh-xxxx xxxxx jane@example"
  }

  # Control Plane Settings
  # ----------------------
  # Example s3 configuration:
  # S3 documentation  https://docs.k3s.io/cli/server
[...]

PR #1034, opened 2024-09-26
Remove the DCO check for providers, and modules entry submissions

cam72cam:
The DCO signoff of providers/modules was added as a precaution in the early days of the OpenTofu registry.  Upon further analysis we have determined that, given the contents of the submission, it is not applicable and are removing it.


Yantrio (approved):

ghost:
Related to #836

ghost (approved):

PR #1082, opened 2024-10-01
Add support for provider specific keys

cam72cam:
We now scan both ./keys/n/namespace/ and ./keys/n/namespace/provider/. This allows users to provide both organization level keys and provider specific keys.

### DIFF
--- a/.github/ISSUE_TEMPLATE/provider_key.yml
+++ b/.github/ISSUE_TEMPLATE/provider_key.yml
@@ -10,6 +10,13 @@ body:
       description: GitHub Username or Organization that contains Providers
     validations:
       required: true
+  - type: input
+    id: providername
+    attributes:
+      label: Provider Name
+      description: Name of the provider this key is used for (without the terraform-provider- prefix). If omitted, this key will be added at the namespace/organization level.
+    validations:
+      required: false
   - type: checkboxes
     id: public_membership
     attributes:
--- a/.github/ISSUE_TEMPLATE/provider_key.yml
+++ b/.github/ISSUE_TEMPLATE/provider_key.yml
@@ -13,7 +13,7 @@ body:
   - type: input
     id: providername
     attributes:
-      label: Provider Name
+      label: Provider Name (Optional)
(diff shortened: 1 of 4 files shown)


src/internal/v1api/providers.go
```diff
@@ -84,8 +84,9 @@ func (p ProviderGenerator) VersionDetails() (map[string]ProviderVersionDetails,
 	versionDetails := make(map[string]ProviderVersionDetails)
 
 	keyCollection := gpg.KeyCollection{
-		Namespace: p.Provider.EffectiveNamespace(),
-		Directory: p.KeyLocation,
+		Namespace:    p.Provider.EffectiveNamespace(),
+		ProviderName: p.Provider.ProviderName,
+		Directory:    p.KeyLocation,
```
ghost, line 89: I think this should be called "name" or "reponame" depending on what it contains. Also, this should be validated to avoid problems.

ghost (dismissed):

src/internal/v1api/providers.go
ghost: Additionally, this will break BC because older OpenTofu and Terraform versions will simply ignore this field.

cam72cam: This is the standard naming convention within this codebase.  Additionally, it is never exposed to users directly.  This step gathers the keys and then exposes them in the expected fashion in the json annotated generation structs.  That part has not changed.

Please go back to bed and get better :heart:

ghost (commented):
Maybe we should extend the provider submission form with a GPG key field (optional)?

.github/ISSUE_TEMPLATE/provider_key.yml
```diff
@@ -10,6 +10,13 @@ body:
       description: GitHub Username or Organization that contains Providers
     validations:
       required: true
+  - type: input
+    id: providername
+    attributes:
+      label: Provider Name
```
Yantrio:
```suggestion
      label: Provider Name (Optional)
```

Yantrio (approved):
LGTM, Just that one tiny nitpick 👍

PR #1123, opened 2024-10-11
Fixes #410: Add provider warnings

### LINKED ISSUE #410: Add the ability to add provider warnings

`tofu` understands that if the .Warnings field is populated when fetching metadata about a provider that it should display those warnings to the end user.

Right now we don't have a method to add a warning to a provider or a specific provider version in this registry. We should allow people to fill in a form for a whole provider, or just a provider version to add warnings that will be displayed when users try to install a provider.

### LINKED ISSUE #751: Deprecate maas/terraform-provider-maas
As requested in #739 by @troyanov, the maas/terraform-provider-maas should be deprecated in favor of canonical/terraform-provider-maas. (Related to #410)

```
╷
│ Warning: Additional provider information from registry
│
│ The remote registry returned warnings for maas/maas:
│ - This provider has moved to canonical/maas. Please update your source in required_providers.
╵
```

_Originally posted by @troyanov in https://github.com/opentofu/registry/issues/739#issuecomment-2248341992_

ghost:
This PR adds the ability to include warnings for providers. This will translate to the (undocumented) "warnings" field in the V1 API response, which [OpenTofu processes](https://github.com/opentofu/opentofu/blob/7fddff6ba4f6e2338361fcd9b33587eae6d09709/internal/getproviders/registry_client.go#L138). This PR also marks all HashiCorp providers that are archived as deprecated, indicating an official replacement where found/available.

Fixes #410: Add provider warnings
Fixes #751: Deprecate maas/terraform-provider-maas
Related to: opentofu/libregistry#39

Example output in the `versions` response:

```json5
{
  "versions": [
    //...
  ],
  "warnings": [
    "This provider is deprecated."
  ]
}
```

### DIFF
--- a/.github/workflows/validate-json-provider.yml
+++ b/.github/workflows/validate-json-provider.yml
@@ -39,7 +39,4 @@ jobs:
         env:
           CHANGED_FILES: ${{ steps.updated.outputs.all_changed_files }}
         run: |
-          for path in "$CHANGED_FILES"
-          do
-            /tmp/validate/run provider "$path"
-          done
+          /tmp/validate/run provider "$CHANGED_FILES"
--- a/providers/a/a10networks/vthunder.json
+++ b/providers/a/a10networks/vthunder.json
@@ -1,4 +1,7 @@
 {
+  "warnings": [
+    "This provider is deprecated. Please use a10networks/thunder instead."
+  ],
   "versions": [
     {
       "version": "0.4.6",
--- a/providers/h/hashicorp/arukas.json
+++ b/providers/h/hashicorp/arukas.json
@@ -1,4 +1,7 @@
 {
+  "warnings": [
+    "This provider is deprecated."
+  ],
   "versions": [
     {
       "version": "1.1.0",
--- a/providers/h/hashicorp/auth0.json
+++ b/providers/h/hashicorp/auth0.json
@@ -1,4 +1,7 @@
 {
+  "warnings": [
+    "This provider is deprecated. Please use auth0/auth0 instead."
+  ],
   "versions": [
     {
       "version": "0.14.0",
--- a/providers/h/hashicorp/azure-classic.json
+++ b/providers/h/hashicorp/azure-classic.json
@@ -1,4 +1,7 @@
 {
+  "warnings": [
+    "This provider is deprecated. Please use hashicorp/azurerm instead."
+  ],
   "versions": [
     {
       "version": "0.1.1",
--- a/providers/h/hashicorp/bitbucket.json
+++ b/providers/h/hashicorp/bitbucket.json
@@ -1,4 +1,7 @@
 {
+  "warnings": [
+    "This provider is deprecated."
+  ],
   "versions": [
     {
       "version": "1.2.0",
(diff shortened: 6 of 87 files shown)


Yantrio (approved):

Yantrio:
I'll approve but im unsure about this failing check. It looks broken to me. can you look into it please @abstractionfactory ?

ghost:
@Yantrio just checked, that one is indeed broken and should be removed. It tries to validate the `/versions` endpoint against a schema for the individual versions. Safe to ignore.

tom-reinders:
@abstractionfactory Is not just the github action broken for multiple files?

This https://github.com/opentofu/registry/actions/runs/11287627799/workflow?pr=1123#L42 should just be 
```
for path in ${CHANGED_FILES}; do
```
as per examples in actions README https://github.com/tj-actions/changed-files?tab=readme-ov-file#using-local-git-directory-

ghost:
@tom-reinders I don't believe so, I checked the verification application and it doesn't actually work when passed a versions endpoint file.

tom-reinders:
But the error in the action is trying to open all the files at ones in https://github.com/opentofu/registry/blob/main/src/cmd/validate/main.go#L93.
The only way for the validate command to get all path at ones in stead of seperated by the foreach is if the foreach is broken and it is.

tom-reinders:
This
```
          for path in "$CHANGED_FILES"
          do
            /tmp/validate/run provider "$path"
          done
```
Does the same as this
```
/tmp/validate/run provider "$CHANGED_FILES"
```

Anything in between the `"` is 1 item in the foreach `for path in "one path" "second path"` and because the action is putting the full content of `CHANGED_FILES` in between the first set of `"` it sees it as one path.

ghost:
So it's broken in more than one place. Wonderful. :D

ollevche:
Hey @abstractionfactory, it seems like the validate command doesn't know how to proceed with multiple files. Thus it fails to open multiple files in this line: https://github.com/opentofu/registry/blob/main/src/cmd/validate/main.go#L62 . Should be pretty easy to fix

ollevche (approved):

ollevche:
GitHub doesn't re-trigger provider validation because the latest commit has no changes to provider paths. Ideally, we would want to show final checks for the diff with the main (or other branch to merge to), because currently it is the way to bypass required validations.

Here is the throwaway commit to validate the check run just fine for this particular PR: https://github.com/opentofu/registry/actions/runs/11326232308/job/31494751924

tom-reinders:
If you update the command to accept multiple paths than also update the command call in the GitHub action from
```
for path in "$CHANGED_FILES"
do
  /tmp/validate/run provider "$path"
done
```
to
```
/tmp/validate/run provider "$CHANGED_FILES"
```
to prevent people from making the mistake of thinking that the for loop really works what it does not.

ollevche:
@tom-reinders makes sense! Here is the final run check: https://github.com/opentofu/registry/actions/runs/11326706265/job/31496187527

cam72cam (approved):

tom-reinders (approved):
LGTM!

PR #1135, opened 2024-10-14
Fixing shellcheck

ghost:


ollevche:
~~Do we also want to skip validation if there is no response on optional fields?~~

https://github.com/opentofu/registry/issues/1118#issuecomment-2410628664

ghost:
@ollevche this is a check for the shell scripts.

ollevche (approved):

PR #1249, opened 2024-11-04
Remove terraform-provider-gridscale 1.27.0 from gridscale.json

nvthongswansea:
Gridscale updated its GPG key so the entries for 1.27.0 is not correct anymore.


ghost:
To clarify, this seems to be a checksum mismatch rather than a GPG key issue. It looks like provider version 1.27.0 was re-released and the checksums in the registry don't match the SHASUMS file currently published. I'm merging this PR to let the update process re-index this version.

PR #1330, opened 2024-11-26
Delete tdabasinskas/backstage provider

tdabasinskas:
This provider was superseded by [`datolabs-io/backstage`](https://search.opentofu.org/provider/datolabs-io/backstage/latest).

### DIFF
--- a/providers/t/tdabasinskas/backstage.json
+++ b/providers/t/tdabasinskas/backstage.json
@@ -1,4 +1,7 @@
 {
+  "warnings": [
+    "This provider is deprecated. Please use datolabs-io/backstage instead."
+  ],
   "versions": [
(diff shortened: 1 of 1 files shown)


ghost:
Hi @tdabasinskas thank you for this PR. Generally we prefer to add a warning to the provider so users of it won't be met with a confusing error message but are instead directed to the correct place. See [this example](https://github.com/opentofu/registry/blob/beac92753ddad1eb16adb2d743bc9c2d761465a9/providers/h/hashicorp/incapsula.json#L3). Would this be ok with you?

tdabasinskas:
Yeah, makes sense - updated.

ghost (approved):

PR #1423, opened 2025-01-04
feat: validate gpg key against existing list of providers

### LINKED ISSUE #356: Validate uploaded GPG keys against providers

As we've seen in #352, providers can upload the wrong key for providers.  We should validate that they key matches the already existing providers in the add key workflow.

diofeher:
Closes #356 

Depends on https://github.com/opentofu/libregistry/pull/47

Code based on @cam72cam's [work](/opentofu/registry/tree/provider_key_checker)

## Description

This code is used to verify if the provided key by the user is really signing the packages they meant to be signing. There were multiple instances when the user provided wrong keys and then commits had to be reverted in order to provide the right signing key.

Failing signature:
Screenshot 2025-01-10 at 12 44 55

Working signature:
Screenshot 2025-01-10 at 12 54 17


ghost:
Thanks @diofeher I'm out for the weekend, I'll review first thing on Monday!

diofeher:
Thanks @abstractionfactory, enjoy your weekend!

src/cmd/verify-gpg-key/main.go
```diff
@@ -36,19 +40,17 @@ func main() {
 		os.Exit(1)
 	}
 
-	ctx := context.Background()
+	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
```
ghost: This timeout may be too short given how slow GitHub Actions can sometimes be and that we now need to also download a provider binary. I'd configure at least a minute.

src/cmd/verify-gpg-key/verify-key-in-providers.go
```diff
@@ -0,0 +1,37 @@
...
+	"time"
+
+	"github.com/opentofu/libregistry/metadata"
+	"github.com/opentofu/libregistry/provider_verifier"
+	"github.com/opentofu/libregistry/types/provider"
+)
+
+func buildKeyVerifier(dataAPI metadata.API) (provider_verifier.KeyVerification, error) {
+	httpClient := http.Client{
+		Timeout: time.Second * 10,
```
ghost: As before, I believe this timeout is too low.

src/cmd/verify-gpg-key/verify-key-in-providers.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func listProviders(ctx context.Context, dataAPI metadata.API, namespace string) ([]provider.Addr, error) {
+	providers, err := dataAPI.ListProvidersByNamespace(ctx, namespace, false)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(providers) == 0 {
+		return nil, fmt.Errorf("no providers found for namespace %s", namespace)
```
ghost:
```suggestion
		return nil, fmt.Errorf("there are no providers in namespace %s; please submit at least one provider before submitting a GPG key", namespace)
```

src/cmd/verify-gpg-key/verify-key-in-providers.go
```diff
@@ -0,0 +1,37 @@
...
+
+	keyVerification, err := provider_verifier.New(httpClient, dataAPI)
+	if err != nil {
+		return nil, err
+	}
+	return keyVerification, nil
+}
+
+func listProviders(ctx context.Context, dataAPI metadata.API, namespace string) ([]provider.Addr, error) {
+	providers, err := dataAPI.ListProvidersByNamespace(ctx, namespace, false)
```
ghost: This doesn't take into account the provider-scoped keys we added recently, see #1082. The submitted key needs to be verified against all providers or against a single provider, depending on the scope in the submitting issue.

ghost (changes requested):
Thank you for your work @diofeher !

src/cmd/verify-gpg-key/main.go
diofeher: Makes sense, gonna change for both.

ghost: Note: the cancel function needs to be called in all cases to not leak goroutines.

diofeher: https://github.com/opentofu/registry/pull/1423/commits/e1faecf18fce0c4212ef88e453664b292912965e fixed.

src/cmd/verify-gpg-key/verify-key-in-providers.go
diofeher: Fixed on https://github.com/opentofu/registry/pull/1423/commits/b18b1f23ed8f74ee0d92e88ba0903c0f7caecb3a

ghost (approved):
Looks good, I'd like someone else familiar with the registry to also take a look.

src/cmd/verify-gpg-key/main.go
```diff
@@ -8,22 +8,31 @@ import (
...
 	"github.com/opentofu/registry-stable/internal/files"
 	"github.com/opentofu/registry-stable/internal/github"
 	"github.com/opentofu/registry-stable/internal/gpg"
 	"github.com/opentofu/registry-stable/pkg/verification"
 )
 
 func main() {
-	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
+	opts := &slog.HandlerOptions{
+		Level: slog.LevelError,
```
cam72cam: Why change the default level to error here?

.github/scripts/submit-provider-key.sh
```diff
@@ -41,7 +41,7 @@ if [[ -n "${providername}" ]]; then
 fi
 
 set +e
-go run ./cmd/verify-gpg-key -org "${namespace}" -username "${GH_USER}" -key-file=tmp.key -output=./output.json
+go run ./cmd/verify-gpg-key -org "${namespace}" -provider-name "${providername}" -username "${GH_USER}" -key-file=tmp.key -output=./output.json
```
ollevche: should we pass `provider-data` flag here as well?

src/cmd/verify-gpg-key/main.go
diofeher: I think I was testing and forgot to remove. Thanks for the catch!

.github/scripts/submit-provider-key.sh
diofeher: thanks, I've put the parameter here as well too.

src/cmd/verify-gpg-key/main.go
```diff
@@ -170,6 +176,38 @@ func VerifyKey(location string) *verification.Step {
...
+	}
+
+	keyVerification, err := buildKeyVerifier(dataAPI)
+	if err != nil {
+		verifyStep.AddError(fmt.Errorf("failed to build key verifier: %w", err))
+		verifyStep.Status = verification.StatusFailure
+		return verifyStep
+	}
+
+	for _, provider := range providers {
```
cam72cam: The interesting part here is that the key really only needs to verify at least one of the providers, not all of them

cam72cam: We could add a summary like: Key is valid for 3/4 applicable providers

Could also break it down by version as well

diofeher: So, what should happen if any of the verifications fail? Just skip it and test the next one?

diofeher: If you have a preferred output for the version, please let me know and I can implement it :) @cam72cam

cam72cam: Yep, just skip the failing one.  We need at least one to pass for this to be considered a valid key.  I don't have a strong preference on format, but we should convey how many versions for each provider this key is valid for.

diofeher: Screenshot 2025-01-29 at 09 35 39

diofeher: Fixed on https://github.com/opentofu/registry/pull/1423/commits/7bc61df3afeef785b15fbf57602e29fb13bec88b

ghost:
Moving back to draft since the PR in libregistry isn't merged yet. We will review this PR fully once the libregistry one is merged.

cam72cam:
With permission from @diofeher, I've modified this branch to coalesce the changes in libregistry and this repo, as well as make it match OpenTofu's understanding of provider keys.

yottta (approved):
Thanks for this. Definitely nicer!
Today I learned something new. Was not aware of this `openpgp.CheckDetachedSignature`. 🤯

PR #1439, opened 2025-01-13
Better error handling when SHASUM file is missing

### LINKED ISSUE #280: Better error message for missing SHA256SUMS file

Currently, a missing SHA256SUMS file causes the following error message:

> No versions detected for repository {URL}

This is not helpful and does not tell the user what to do. The error message should be more helpful, point to the GitHub Action run and/or contain the specific error cause.

### OpenTofu Version

```shell
N/A
```

N/A

diofeher:
Closes #280 

# Description

- Removed a few extra spaces on the texts
- Add output.json as default on this script 
- Removed an extra slash on the github-action bot comment (https://github.com/opentofu/registry/pull/1438) after the src.
- Provided a better error message when shaSum file is missing.

Tested with https://github.com/diofeher/terraform-provider-nix-signature/releases/tag/v0.1.1

Screenshot 2025-01-13 at 19 04 00

### DIFF
--- a/src/cmd/add-provider/main.go
+++ b/src/cmd/add-provider/main.go
@@ -28,7 +28,7 @@ func main() {
 	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
 
 	repository := flag.String("repository", "", "The provider repository to add")
-	outputFile := flag.String("output", "", "Path to write JSON result to")
+	outputFile := flag.String("output", "output.json", "Path to write JSON result to")
 	providerDataDir := flag.String("provider-data", "../providers", "Directory containing the provider data")
 
 	flag.Parse()
@@ -70,8 +70,8 @@ func main() {
 		}
 		for _, p := range providers {
 			if strings.ToLower(p.RepositoryURL()) == strings.ToLower(submitted.RepositoryURL()) {
-				output.Exists = true
-				return fmt.Errorf("Repository already exists in the registry, %s", p.RepositoryURL())
+				// 	output.Exists = true
+				// 	return fmt.Errorf("Repository already exists in the registry, %s", p.RepositoryURL())
 			}
 		}
 
--- a/src/cmd/add-provider/main.go
+++ b/src/cmd/add-provider/main.go
@@ -70,8 +70,8 @@ func main() {
 		}
 		for _, p := range providers {
 			if strings.ToLower(p.RepositoryURL()) == strings.ToLower(submitted.RepositoryURL()) {
-				// 	output.Exists = true
-				// 	return fmt.Errorf("Repository already exists in the registry, %s", p.RepositoryURL())
+				output.Exists = true
+				return fmt.Errorf("Repository already exists in the registry, %s", p.RepositoryURL())
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


ghost:
This PR seems to have commented-out code, moving back to draft.

ghost (commented):
I'm not sure if it's possible, but it would be nice if this functionality reported back as a comment to the submitter because it would make debugging issues easier. Then again, most of this repository would need to be moved over to libregistry anyway, so maybe we can do it as part of that move.

src/cmd/add-provider/main.go
```diff
@@ -28,7 +28,7 @@ func main() {
 	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
 
 	repository := flag.String("repository", "", "The provider repository to add")
-	outputFile := flag.String("output", "", "Path to write JSON result to")
+	outputFile := flag.String("output", "output.json", "Path to write JSON result to")
```
diofeher, line 31: It's just because if you don't specify the output flag, the program panics next to the JSON function that writes to this output.

diofeher: So I'm adding the default in case someone doesn't pass the flag

diofeher:
Your review makes sense @abstractionfactory ! I will work later to port this code to libregistry, and add the code to report back as a comment in the registry repo. I'll leave this PR open for now for changing the Github actions to report the comment.

ghost:
@diofeher I think we can skip the libregistry part for now, we still need to figure out what our roadmap is to get that going. However, I'd still like to get this merged before that.

src/cmd/add-provider/main.go
cam72cam: I'd rather the program exit early if the flag is not provided

src/internal/provider/shasum.go
```diff
@@ -13,7 +13,7 @@ func (p Provider) GetSHASums(shaFileDownloadUrl string) (map[string]string, erro
 		return nil, fmt.Errorf("failed to download asset contents: %w", assetErr)
 	}
 	if contents == nil {
-		return nil, nil
+		return nil, fmt.Errorf("SHASUM file is empty or missing: %s", shaFileDownloadUrl)
```
cam72cam, line 16: There was a very specific reason I didn't return an error here.  We don't want new releases (with missing shasums) to cause a whole provider to stop updating.  A lot of providers have historical releases that we scan from time to time which would percolate this error back up the stack.

We either need to add a flag to the calls tack to make this a true error or not, or create a specific error type that is handled correctly.

diofeher: Makes sense @cam72cam 

Should we put what your explained to me in this part of the code as well? I feel this is important information that should be in a comment

diofeher: So this seems to be more complicated than I thought, because all workflows are using this function, so we would need to identify what exact behavior we need when adding a provider from Github Actions or from other places. I'm gonna do a research later on this.

cam72cam: A comment is a great idea, a lot of this was hastily adapted from prototype so context may be missing for quirks like this.

PR #1487, opened 2025-01-29
fix: warning error for slog

diofeher:
When running tests:

```
cmd/run-server/main.go:21:41: slog.Logger.Error arg "err" should be a string or a slog.Attr (possible missing key or value)
cmd/run-server/main.go:32:42: slog.Logger.Error arg "err" should be a string or a slog.Attr (possible missing key or value)
```

Fixed this.


ghost (approved):

cam72cam:
Thanks @diofeher !

PR #1570, opened 2025-02-18
fix: lowercase for provider naming

### LINKED ISSUE #1528: Provider with a version that contains uppercase characters cannot be fetched
> NOTE: Initially this issue was named `Provider with a pre-release version cannot be fetched` but upon investigation by the team the title has been updated to match the actual issue.

In the case of the `zededa/zedcloud` provider the *latest* version is `2.3.1-RC1` .

That is correctly displayed here: https://search.opentofu.org/provider/zededa/zedcloud/latest and all the metadata information seems to be correctly updated in `providers/z/zededa/zedcloud.json`.

In `versions` it is also reported correctly:
```
❯ curl https://registry.opentofu.org/v1/providers/zededa/zedcloud/versions | jq .
{
  "versions": [
    {
      "version": "2.3.1-RC1",
      "protocols": [
        "5.0"
      ],
      "platforms": [
        {
          "os": "darwin",
          "arch": "amd64"
        },
........................................
        {
          "os": "linux",
          "arch": "amd64"
        },
........................................
```

But trying to get the download information results in a `404` and some Cloudflare HTML:
```
curl -w "HTTP status code: %{http_code}\n" https://registry.opentofu.org/v1/providers/zededa/zedcloud/2.3.1-RC1/download/linux/amd64
<!DOCTYPE html>
<html lang="en">
  <head>
........................................
  </body>
</html>
HTTP status code: 404
```

```
❯ tofu init

Initializing the backend...

Initializing provider plugins...
- Finding zededa/zedcloud versions matching "2.3.1-RC1"...
╷
│ Error: Incompatible provider version
│
│ Provider registry.opentofu.org/zededa/zedcloud v2.3.1-RC1 does not have a package available for your current platform, linux_amd64.
│
│ Provider releases are separate from OpenTofu CLI releases, so not all providers are available for all platforms. Other versions of this provider may have different platforms supported.
╵
```

### OpenTofu Version

```shell
OpenTofu v1.9.0
on linux_amd64
```

```hcl
terraform {
  required_providers {
    zedcloud = {
      source  = "zededa/zedcloud"
      version = "2.3.1-RC1"
    }
  }
}
```

diofeher:
Resolves https://github.com/opentofu/registry/issues/1528 by adding transformation logic into version strings, lowering cases, but maintaining its case when referring to Github URLs and other stuff that shouldn't be transformed.

Added a couple of tests to guarantee is working in all edge cases.


src/internal/v1api/providers_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"testing"
+
+	"github.com/opentofu/registry-stable/internal/github"
+	"github.com/opentofu/registry-stable/internal/provider"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func newTestServer(t testing.TB) *httptest.Server {
```
yottta: `question`
Shouldn't hurt, but is it really necessary to have `testing.TB` here instead of just `*testing.T`?

yottta (commented):
LGTM. I would still like a review from somebody with more knowledge on the registry side.

src/internal/v1api/providers_test.go
diofeher: Whoa, good catch! When I was reading about these t.Helper, they were mentioned about using t testing.TB, so I didn't mind to test if it was going to work using t *testing.T. Thanks for the learning here!

src/internal/provider/version.go
```diff
@@ -27,25 +28,44 @@ var (
 	}
 )
 
+type VersionFromTagArgs struct {
+	Release   string
+	URLPrefix string
```
cam72cam: Why choose this point to add URLPrefix?  This entire codebase assumes that the origin is github.  Changing that is a massive undertaking.

ollevche:
Given that we lowercase all the incoming requests to R2, I believe we can limit the fix to cover file generation only. 

I am not extra familiar with the codebase, but I checked the `cmd/generate-v1` command and it uses `ProviderGenerator.Generate` and `ModuleGenerator.Generate` to prepare files for R2 upload:

https://github.com/opentofu/registry/blob/086be00e0df46ac595eb47808652198f12689531/src/cmd/generate-v1/main.go#L55-L66

The resulting paths are generated via `ProviderGenerator.VersionDownloadPath` and `ModuleGenerator.VersionDownloadPath`, so I believe tweaking those functions would do the trick. We could change the lowercasing of version here:

https://github.com/opentofu/registry/blob/086be00e0df46ac595eb47808652198f12689531/src/internal/v1api/providers.go#L53-L57

We don't need to change the metadata itself and we only focus on file upload. What do you think about this approach?

diofeher:
Basically was just to be able to test. I'm passing the URL from the http
test server

Diógenes Fernandes
http://diofeher.net/

src/internal/provider/version.go
diofeher: Basically this change was introduced in order to be able to test. I'm passing the URL from the http test server

ollevche (approved):
Looks good!

src/internal/v1api/providers.go
```diff
@@ -53,7 +53,8 @@ func (p ProviderGenerator) VersionListingPath() string {
 func (p ProviderGenerator) VersionDownloadPath(ver provider.Version, details ProviderVersionDetails) string {
 	namespacePath := strings.ToLower(p.Provider.Namespace)
 	providerNamePath := strings.ToLower(p.Provider.ProviderName)
```
apparentlymart, line 55: I realize that this is a separate concern from what you were trying to fix here and so it'd be reasonable to say you don't want to change it in this PR, but FWIW we have [`tfaddr.ParseProviderPart`](https://pkg.go.dev/github.com/opentofu/registry-address#ParseProviderPart) as an opinionated function for validating and normalizing these segments of a provider source address.

Using that function would make the registry normalize these names the same way OpenTofu itself does.

(We don't really have a similar function for _version numbers_ because for that OpenTofu follows semver which treats the prerelease and build fields as case-sensitive, and so our choice to lowercase those here is a registry policy decision rather than a true part of the OpenTofu provider protocol. But that's fine as long as the registry is consistent in the normalization it does, and as long as users write version constraints that match the normalization.)

src/internal/v1api/modules_test.go
```diff
@@ -25,6 +25,11 @@ func Test_ModuleGenerator(t *testing.T) {
 	v := module.Version{
 		Version: "v1.0.1",
 	}
+
+	vl := module.Version{
+		Version: "v2.3.1-RC1",
+	}
 	assert.Equal(t, "gen/v1/modules/spacename/name/target/versions", m.VersionListingPath())
 	assert.Equal(t, "gen/v1/modules/spacename/name/target/1.0.1/download", m.VersionDownloadPath(v))
+	assert.Equal(t, "gen/v1/modules/spacename/name/target/2.3.1-rc1/download", m.VersionDownloadPath(vl))
```
cam72cam, line 34: I'm expecting the opposite here, we should not be lowercasing version numbers at all.

ollevche: Currently, we don't lowercase it before upload and the problem arises when users try to get that file from R2, where all the incoming requests got lowercased. The easiest solution would be to upload them in lowercase (as we do with other path parts). What do I miss?

src/internal/v1api/providers.go
diofeher: That's a good point @apparentlymart 

I'm not sure if we would want to use it here though because we're dealing with different rules, for example, we're using a rewrite rule on Cloudflare that lowercase all incoming requests, so it would be different from what we have at OpenTofu.

src/internal/v1api/modules_test.go
diofeher: @ollevche is right, the idea is that these files are uppercased on the R2 bucket and then we receive the request and the path is rewrote to be lowercase, the files are not found.

cam72cam: Ah, thanks for the clarification!

cam72cam (approved):

cam72cam:
I've identified that there are a number of other providers impacted by this issue:
```
find ../generated/ -maxdepth 5 | grep '[A-Z]'
../generated/v1/providers/c2devel/croccloud/4.14.0-CROC4
../generated/v1/providers/c2devel/croccloud/4.14.0-CROC6
../generated/v1/providers/c2devel/croccloud/4.14.0-CROC7
../generated/v1/providers/c2devel/croccloud/4.14.0-CROC8
../generated/v1/providers/c2devel/croccloud/4.14.0-CROC1
../generated/v1/providers/c2devel/croccloud/4.14.0-CROC5
../generated/v1/providers/c2devel/croccloud/4.14.0-CROC10
../generated/v1/providers/c2devel/croccloud/4.14.0-CROC3
../generated/v1/providers/c2devel/croccloud/4.14.0-CROC9
../generated/v1/providers/c2devel/croccloud/4.14.0-CROC2
../generated/v1/providers/cloudstack/cloudstack/0.4.0-RC1
../generated/v1/providers/ciscodevnet/cdo/0.12.6-LH-70847-fix-release-trigger.1
../generated/v1/providers/grixan482/grixan482/4.14.0-GRIXAN1
../generated/v1/providers/utilitywarehouse/nebraska/0.1.4-RC1
../generated/v1/providers/utilitywarehouse/nebraska/0.1.3-RC2
../generated/v1/providers/utilitywarehouse/nebraska/0.1.3-RC1
../generated/v1/providers/utilitywarehouse/nebraska/0.1.5-RC3
../generated/v1/providers/utilitywarehouse/nebraska/0.1.5-RC4
../generated/v1/providers/utilitywarehouse/nebraska/0.1.6-RC1
../generated/v1/providers/samuelchong/ddcloud/3.0.7-Fancy
../generated/v1/providers/zededa/zedcloud/2.3.1-RC1
../generated/v1/providers/zededa/zedcloud/2.3.0-RC
../generated/v1/providers/pixarv/r4/24.2.2-RITT0
../generated/v1/providers/pixarv/ritt/4.14.0-RITT3
../generated/v1/providers/pixarv/ritt/4.14.1-RITT2
../generated/v1/providers/pixarv/ritt/4.14.1-RITT29
../generated/v1/providers/pixarv/ritt/4.14.1-RITT33
../generated/v1/providers/pixarv/ritt/4.14.0-RICO1
../generated/v1/providers/pixarv/ritt/4.14.1-RITT36
../generated/v1/providers/pixarv/ritt/4.14.1-RITT34
../generated/v1/providers/pixarv/ritt/4.13.0-RITT0
../generated/v1/providers/pixarv/ritt/4.14.1-RITT3
../generated/v1/providers/pixarv/ritt/4.14.0-RICO0
../generated/v1/providers/pixarv/ritt/4.13.0-RITT1
../generated/v1/providers/pixarv/ritt/4.14.0-RITT1
../generated/v1/providers/pixarv/ritt/4.14.1-RITT17
../generated/v1/providers/pixarv/ritt/4.14.1-RITT13
../generated/v1/providers/pixarv/ritt/4.14.0-RICO2
../generated/v1/providers/pixarv/ritt/4.14.1-RITT31
../generated/v1/providers/pixarv/ritt/4.14.1-RITT4
../generated/v1/providers/pixarv/ritt/1.14.0-RITT5
../generated/v1/providers/pixarv/ritt/4.14.0-RITT11
../generated/v1/providers/pixarv/ritt/1.14.0-RITT6
../generated/v1/providers/pixarv/ritt/4.14.1-RITT37
../generated/v1/providers/pixarv/ritt/4.14.1-RITT24
../generated/v1/providers/pixarv/ritt/4.14.1-RITT18
../generated/v1/providers/pixarv/ritt/4.14.1-RITT19
../generated/v1/providers/pixarv/ritt/4.14.0-RITT10
../generated/v1/providers/pixarv/ritt/4.14.1-RITT28
../generated/v1/providers/pixarv/ritt/4.14.1-RITT22
../generated/v1/providers/pixarv/ritt/4.14.0-RITT9
../generated/v1/providers/pixarv/ritt/4.14.1-RITT23
../generated/v1/providers/pixarv/ritt/4.14.1-RITT25
../generated/v1/providers/pixarv/ritt/4.14.1-RITT38
../generated/v1/providers/pixarv/ritt/4.14.1-RITT27
../generated/v1/providers/pixarv/ritt/4.14.1-RITT20
../generated/v1/providers/pixarv/ritt/4.14.1-RITT26
../generated/v1/providers/pixarv/ritt/4.14.1-RITT12
../generated/v1/providers/pixarv/ritt/4.14.1-RITT11
../generated/v1/providers/pixarv/ritt/4.14.0-RITT8
../generated/v1/providers/pixarv/ritt/4.9.0-RITT0
../generated/v1/providers/pixarv/ritt/4.14.1-RITT30
../generated/v1/providers/pixarv/ritt/4.14.1-RITT15
../generated/v1/providers/pixarv/ritt/4.14.0-RITT2
../generated/v1/providers/pixarv/ritt/4.14.1-RITT7
../generated/v1/providers/pixarv/ritt/4.14.1-RITT5
../generated/v1/providers/pixarv/ritt/4.14.0-RITT7
../generated/v1/providers/pixarv/ritt/4.14.1-RITT32
../generated/v1/providers/pixarv/ritt/4.14.1-RITT21
../generated/v1/providers/pixarv/ritt/4.14.1-RITT16
../generated/v1/providers/pixarv/ritt/4.14.1-RITT6
../generated/v1/providers/pixarv/ritt/4.14.1-RITT10
../generated/v1/providers/pixarv/ritt/4.14.1-RITT9
../generated/v1/providers/pixarv/ritt/4.13.0-RITT2
../generated/v1/providers/pixarv/ritt/4.14.1-RITT35
../generated/v1/providers/pixarv/ritt/4.14.1-RITT40
../generated/v1/providers/pixarv/ritt/4.14.0-RITT0
../generated/v1/providers/pixarv/ritt/4.9.0-RITT1
../generated/v1/providers/pixarv/ritt/4.14.1-RITT39
../generated/v1/providers/pixarv/ritt/4.14.1-RITT8
../generated/v1/providers/pixarv/ritt/4.14.1-RITT14
../generated/v1/providers/anotherchrisk/nsxt/3.2.7-DEV
```

I can run the generate-all from github actions once this is merged

src/internal/v1api/providers.go
apparentlymart: Fair enough!

I think we still ought to be careful here though: there are lots of different ways to implement "convert to lowercase" with support for different subsets of the world languages and so I would also want to confirm that what Cloudflare is doing matches what Go's ToLower does. 

The algorithm used by OpenTofu's own source address parser is based on an IETF standard (the nameprep profile of stringprep) which includes a specific opinion on how to normalize to lowercase, so we have now potentially three different interpretations of "convert to lowercase" here. Though I suppose another option to deal with it is with a policy that OpenTofu Registry only allows ASCII characters in names, which is a restrictive subset that I think everyone can agree on what it means to convert to lowercase. I've not seen non-ASCII names in wide use, and we've found that OCI Distribution also only allows ASCII characters and so that restriction would ensure that any provider in our registry can be mapped successfully into an OCI repository acting as a mirror.

(Folks running other registries or using other provider installation methods would still be able to use any character permitted by "nameprep" or course, since that's how the generic registry protocol works.)

apparentlymart: I suppose it's also worth noting that by the time OpenTofu is making requests to the registry it has already applied the nameprep algorithm client-side, and so the registry server will never see a non-normalizer request from OpenTofu CLI itself. Other clients might behave differently, of course.

diofeher: That's a great and thoughtful comment @apparentlymart , I didn't think of that. Do you think is worth it to create a function to normalize these paths to make it easier to maintain or this PR should be good to go?

Something like 

```
func normalizePath(namespace, provider, version string) {
   return strings.toLower(fmt.Sprintf("%s/%s/%s", namespace, provider, version))
}

apparentlymart: Admittedly what I've been discussing here is not really relevant to what you were trying to solve with this PR, so I think it's fine to leave it unchanged in this PR and wait to see if there's actually a problem. If we ever get a request to add a non-ASCII name to the registry then we can discuss this then. 

Sorry for the distraction!

PR #1894, opened 2025-03-21
add one additional validation for the provider keys

### LINKED ISSUE #356: Validate uploaded GPG keys against providers

yottta:
Resolved #356

These are the validations that I personally run every time a new provider key is submitted.
I suggest adding these in the script that validates the submitted keys. This way, we also tackle [this part of the PROCEDURES.md](https://github.com/opentofu/registry/blob/main/PROCEDURES.md#user-complaint-the-provider-is-not-signed-with-a-valid-signing-key).

As also discussed in #356, we could also use `tofu init` for doing so but for the moment that is overkill. In case the current approach does not suffice, we could reiterate later.

### DIFF
--- a/.github/scripts/submit-provider-key.sh
+++ b/.github/scripts/submit-provider-key.sh
@@ -60,12 +60,6 @@ if [[ -z "${providername}" ]]; then
 else
   keyfile="../keys/${namespace:0:1}/${namespace}/${providername}/provider-$(date +%s).asc"
 fi
-if [[ -d "$(dirname "${keyfile}")" ]]; then
-  msg="Updated"
-  #git rm $(dirname $keyfile)/*
-else
-  msg="Created"
-fi
 mkdir -p "$(dirname "${keyfile}")"
 mv tmp.key "${keyfile}"
 
--- a/.github/scripts/submit-provider-key.sh
+++ b/.github/scripts/submit-provider-key.sh
@@ -55,10 +55,12 @@ if [[ "${verification}" != 0 ]]; then
   exit 1
 fi
 
+repo=""
 if [[ -z "${providername}" ]]; then
   keyfile="../keys/${namespace:0:1}/${namespace}/provider-$(date +%s).asc"
 else
   keyfile="../keys/${namespace:0:1}/${namespace}/${providername}/provider-$(date +%s).asc"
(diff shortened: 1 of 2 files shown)


.github/scripts/submit-provider-key.sh
```diff
@@ -94,4 +96,26 @@ git push -u origin "${branch}"
...
 pr=$(gh pr create --title "${TITLE}" --body "${msg} ${keyfile/.././} for provider ${namespace}. Closes #${NUMBER}.") #--assignee opentofu/core-engineers)
 gh issue comment "${NUMBER}" -b "Your submission has been validated and has moved on to the pull request phase (${pr}).  This issue has been locked."
-gh issue lock "${NUMBER}" -r resolved
\ No newline at end of file
+gh issue lock "${NUMBER}" -r resolved
+
+if [[ -n "${repo}" ]]; then
+  apt update && apt install -y gpg
+  # get the latest release of the provider
+  latest_release=$(gh release list --exclude-drafts --exclude-pre-releases --repo "${repo}" -L 1 -O desc --json name -q '.[].name')
```
Yantrio: I wonder if we need to do it against the latest release, or just one of the releases?

I'm concerned someone could be uploading a key for an older version of a provider.

Maybe we just make this a warning and say that it should be manually checked by a human?

yottta: > I wonder if we need to do it against the latest release, or just one of the releases? I'm concerned someone could be uploading a key for an older version of a provider.

You're right, but

> Maybe we just make this a warning and say that it should be manually checked by a human?

I wanted to have this more of a warning, indeed.
Since we (or at least I) run those checks manually, I will do it manually when I see that warning:
```shell
  if ! gpg --verify "${sigfile}" "${shafile}"
  then
    gh issue comment "${NUMBER}" -b "Failed to validate the submitted key against ${repo}@${latest_release}. Could this key be for an older version of the provider?"
  fi
``` 

That's why I added this check after the branch creation and after all of the checks that we are having right now.

Yantrio (approved):

ollevche (approved):

PR #2002, opened 2025-04-12
using ctx instead of multiple checks throughout the function

diofeher:
Refactoring the verify key function, using Go's context to save resources on the parallelized requests.

### DIFF
--- a/src/cmd/verify-gpg-key/main.go
+++ b/src/cmd/verify-gpg-key/main.go
@@ -47,8 +47,10 @@ func main() {
 
 	ctx := context.Background()
 	ghClient := github.NewClient(ctx, logger, token)
+	ctxForVerifier, cancelFunc := context.WithCancel(context.Background())
+	ghListClient := github.NewClient(ctxForVerifier, logger, token)
 
-	providers, err := provider.ListProviders(*providerDataDir, *orgName, logger, ghClient)
+	providers, err := provider.ListProviders(*providerDataDir, *orgName, logger, ghListClient)
 	if err != nil {
 		logger.Error("Failed to list providers", slog.Any("err", err))
 		os.Exit(1)
@@ -66,7 +68,7 @@ func main() {
 
 	result := &verification.Result{}
 
-	s := VerifyKey(*keyFile, filteredProviders)
+	s := VerifyKey(*keyFile, filteredProviders, cancelFunc)
 	result.Steps = append(result.Steps, s)
 
 	s = VerifyGithubUser(ghClient, *username, *orgName)
@@ -110,7 +112,7 @@ func VerifyGithubUser(client github.Client, username string, orgName string) *ve
 
 var gpgNameEmailRegex = regexp.MustCompile(`.*\<(.*)\>`)
 
-func VerifyKey(location string, providers provider.List) *verification.Step {
+func VerifyKey(location string, providers provider.List, cancelFunc context.CancelFunc) *verification.Step {
 	verifyStep := &verification.Step{
 		Name: "Validate GPG key",
 	}
@@ -204,7 +206,6 @@ func VerifyKey(location string, providers provider.List) *verification.Step {
 			}
 
 			foundProviderForKey := false
-			ctx, cancelFunc := context.WithCancel(context.Background())
 
 			err = providers.Parallel(20, func(p provider.Provider) error {
 				meta, err := p.ReadMetadata()
@@ -221,12 +222,12 @@ func VerifyKey(location string, providers provider.List) *verification.Step {
 						logger.Info("Begin version check")
 
 						// Inspired by OpenTofu's getproviders
-						shasumResp, err := p.Github.DownloadAssetContents(ctx, version.SHASumsURL)
+						shasumResp, err := p.Github.DownloadAssetContents(version.SHASumsURL)
 						if err != nil {
 							return err
 						}
 
-						sigResp, err := p.Github.DownloadAssetContents(ctx, version.SHASumsSignatureURL)
+						sigResp, err := p.Github.DownloadAssetContents(version.SHASumsSignatureURL)
 						if err != nil {
 							return err
 						}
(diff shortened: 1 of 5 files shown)


src/internal/github/assets.go
```diff
@@ -1,21 +1,28 @@
...
+	"context"
 	"fmt"
 	"io"
 	"log/slog"
 	"net/http"
 )
 
 // DownloadAssetContents downloads the contents of the asset at the given URL and returns it directly
-func (c Client) DownloadAssetContents(downloadURL string) ([]byte, error) {
+func (c Client) DownloadAssetContents(ctx context.Context, downloadURL string) ([]byte, error) {
```
cam72cam: How does this ctx interact with the c.httpClient's transport context?

diofeher: Great question, I didn't know you could pass the ctx to the transport httpClient, https://github.com/opentofu/registry/blob/main/src/internal/github/client.go#L41

I'm going to do some research on it and will update this convo.

diofeher: From what I've understood on reading https://github.com/opentofu/registry/blob/main/src/internal/github/client.go#L84-L86

All context that we pass to our requests, using this httpClient, are going to be overwritten by the ctx defined on this struct:

https://github.com/opentofu/registry/blob/main/src/internal/github/client.go#L80

I'm not sure if this is a good idea. It's expected that context passed to requests are being used, seems like we're doing magic here by doing that. I would at least check if there's an existing context on the request before setting the transport context to the request. What do you think?

diofeher: The other way to think about that is that we will be using the same ctx for all the operations being done in this client, and I think that can be confusing. Let's say we're using a part of the client to request some stuff, and then another part to request the files being used for verifying the provider. I think it would be expected these two different contexts not being able to interact with each other.

cam72cam: Here's an alternate idea:  We could send the context used to construct the github client into this function and use that instead of creating and passing a new context inside here.

diofeher: @cam72cam what if we want to pass a context related only to this function and not to the whole Github client? I thought about using that approach, but I think we would be trespassing limits between different Client functions.

I think the more specific context (this specific function) should be taking precedence over the more general context (the client's transport one).

diofeher: I've been reading other clients' code and I'm seeing they use context as the first parameter for their functions, examples:

https://github.com/google/go-github/blob/master/github/orgs.go#L156
https://github.com/istio/client-go/blob/64a0de0e4ead45370c64c2ff9073839423027841/cmd/example/client.go#L72

And for me it does make sense (compared to what I've been reading on context) to have it like the first argument per function of the client. Let me know what you think about it.

cam72cam: I think I'd like to keep the scope of these changes as small as possible.  Constructing a github client with a context that can be cancelled seems to be the simplest option from what I can see.

diofeher: Hey @cam72cam it seems we can't use the context on the client indeed for the reason I've stated above:

Screenshot 2025-04-14 at 14 44 50

It seems to directly influence other requests that comes later (so we cannot use cancel on the Github's client ctx).

I'm testing with req.Context() != context.Background() because of https://cs.opensource.google/go/go/+/master:src/net/http/request.go;l=356

cam72cam: You could also create two github structs with two different contexts.

diofeher: Agreed that is possible but I feel we would be propagating two anti-patterns here:

1 - Using one context per the global instance, due to the reasons stated above. The user would be expecting to deal with one context's operations without interfering with other operations (like calling a different function on the Github instance) and it's common to expect context to be passed as the first parameter in this kind of function.

2 - If we create two `Github` instances, we won't be reusing the httpClient, therefore not reusing existing TCP connections https://medium.com/@p408865/til-go-http-client-reuse-connection-5d8c48731dec

diofeher: Created an alternative [implementation](https://github.com/opentofu/registry/pull/2002/commits/dcaf1f485355678fe5f9b6e56b60164a660f89ef) based on your comments, but I still think we should consider the two points above.

src/cmd/verify-gpg-key/main.go
```diff
@@ -217,29 +218,19 @@ func VerifyKey(location string, providers provider.List) *verification.Step {
...
 
 						// Inspired by OpenTofu's getproviders
-
 						shasumResp, err := p.Github.DownloadAssetContents(version.SHASumsURL)
 						if err != nil {
-							return nil
-						}
-						if foundProviderForKey {
-							return nil
+							return err
```
Yantrio, line 227: i still think this should not error, in cases where a version exists but they didnt add the the shasums file artifact

cam72cam: I double checked internal/provider/version.go (and related calls), we do return the error there.  @diofeher fixed a bug in my initial implementation.

diofeher: Leaving as a reference: https://github.com/opentofu/registry/blob/main/src/internal/provider/version.go#L50

@Yantrio we're already skipping errors when the shasums file is not found

src/cmd/verify-gpg-key/main.go
```diff
@@ -59,15 +60,15 @@ func main() {
 		filteredProviders = providers
 	} else {
 		for _, provider := range providers {
-			if strings.ToLower(provider.ProviderName) == strings.ToLower(*providerName) {
+			if strings.EqualFold(provider.ProviderName, *providerName) {
```
diofeher, line 63: This was suggested by VsCode as a cleaner way to do this comparison, that's why I've added it @cam72cam

cam72cam: LGTM

cam72cam (approved):

PR #2007, opened 2025-04-14
Organize ciscodevnet keys

cam72cam:
Customized the verify-gpg-key tool to build a full report for the namespace, for each key. See https://github.com/opentofu/registry/commit/79866c2f6ae57e704c38e3d7ff57c5677386f6f6

A few keys were able to be moved to their own directories, but a bunch of the keys were not actually used in the first place.  There was also a duplicate we didn't catch.

Each commit links to the issue that the key was originally submitted under.


diofeher (approved):
Thanks for this investigation @cam72cam ! Great findings!

One question: what is the purpose of having keys named `provider.asc` and other ones with `provider-<timestamp>.asc`?

Other than that, LGTM!

cam72cam:
provider.asc is one of the older filename conventions.  I kept it to help with parsing through the history.

yottta (approved):
Looks good. Thanks for checking this out.

Yantrio (commented):
So, this looks good to me, but I would love someone from ciscodevnet to determine if this is correct or not.

My concern is that they uploaded keys to be non-provider-specific, and they may be re-using those keys in the future for other projects. 

I think the assumptions made here are valid, I think we should just double check anyway just incase! Are you talking to someone there already @cam72cam ?

cam72cam:
I mentioned this a bit in the comments of https://github.com/opentofu/registry/pull/2007/commits/f756bedfac81992d9aac1478654565243ae248c8.  Last I had heard from the ciscodevnet folks, each provider is in it's own team and it will be tricky to get all of the distinct teams to submit keys.  That commit works around the problem, but could be reverted if that situation changes someday.

PR #2009, opened 2025-04-14
Audit submitted provider keys now that tooling exists

cam72cam:
A few classes of issues are addressed in this PR:
* A bug in our actions at one point resulted in terraform-provider-<name> instead of <name> for key directories
* Unused keys added to the registry
* Keys submitted before individual provider keys were supported need to be moved when multiple providers are in the  namespace.


yottta (approved):
TL;DR: Looks good after testing.

Checked and tested every commit you added and you have a summary below:

- claranet tested the moved keys against the associated provider and works well
- akp tested the key and works well
- tss removal is ok, tested that the remaining key is signing all releases of that provider
- grafana validated that it’s the right path
- gridscale removal confirmed by comparing the content of the removed one with the one remaining in gridscale/gridscale
- komminarlabs removal confirmed. The key cannot sign any release from any provider in that org (4 providers)
- meshcloud removal confirmed that it’s ok since the remaining key can sign all of the releases from meshcloud/terraform-provider-meshstack and there are no other active providers in that org
- spirl removal confirmed by comparing the content of the removed one with the one remaining in spirl/spirl
- trendmicro moving confirmed by confirming that the key indeed is not signing other providers from trendmicro

PR #2208, opened 2025-05-21
add file-based lock

tripti-gh:
Adding a file-based lock to avoid multiple instances of job syncing to R2 simultaneously.


Yantrio:
Hi @tripti-gh , whilst we do appreciate the PR, I feel that this approach wouldn't work at all in github actions. Also I don't believe this is currently an issue at all.

If we were to solve this issue, I think we should use the concurrency feature in github actions to ensure that only one action runs at a time. 

If you feel that this is an issue that needs to be solved, please see our [contributing guide](https://github.com/opentofu/registry/blob/main/CONTRIBUTING.md) for information about raising an issue and discussing items before beginning work.

PR #2404, opened 2025-07-07
Added ability to blacklist a certain version from being fetched

Yantrio:
Hashicorp aws 6.1.0 should be revoked but we had no system to do this in the registry. This adds a simple blacklisting functionality.


src/internal/blacklist/blacklist.go
```diff
@@ -0,0 +1,117 @@
...
+	Modules   []ModuleEntry   `json:"modules"`
+}
+
+var (
+	instance *Blacklist
+	once     sync.Once
+	mu       sync.RWMutex
+)
+
+// Load reads the blacklist from the versions_blacklist.json file
```
cam72cam: I'll be honest, I don't love that this is a global singleton.  The path guessing below is particularly egregious.

That said it does seem to work and I understand not wanting to have to pass the blacklist data around.

Yantrio: That's valid, I've re-worked it.

src/cmd/add-module/main.go
```diff
@@ -35,6 +36,15 @@ func main() {
 	flag.Parse()
 
 	ctx := context.Background()
+
+	bl, err := blacklist.Load()
+	if err != nil {
+		logger.Error("Failed to load blacklist, proceeding without it", slog.Any("err", err))
```
cam72cam: This log message does not make sense.

src/cmd/add-provider/main.go
```diff
@@ -34,6 +35,15 @@ func main() {
 	flag.Parse()
 
 	ctx := context.Background()
+
+	bl, err := blacklist.Load()
+	if err != nil {
+		logger.Error("Failed to load blacklist, proceeding without it", slog.Any("err", err))
```
cam72cam: same as above

src/internal/blacklist/blacklist.go
```diff
@@ -0,0 +1,107 @@
...
+	Name         string `json:"name"`
+	TargetSystem string `json:"target_system"`
+	Version      string `json:"version"`
+	Reason       string `json:"reason"`
+}
+
+type Blacklist struct {
+	Providers []ProviderEntry `json:"providers"`
+	Modules   []ModuleEntry   `json:"modules"`
+	mu        sync.RWMutex
```
cam72cam: Why does this lock still exist?

src/cmd/add-module/main.go
Yantrio: Oops, i moved this from warning to error and forgot to update the error message. my bad!

src/internal/module/build.go
```diff
@@ -27,6 +28,12 @@ func (m Module) UpdateMetadataFile() error {
 // BuildMetadata builds the Metadata for the module by collating the tags from
 // the module repository.
 func (m Module) BuildMetadata() (*Metadata, error) {
+	// Use blacklist from Module struct, fall back to empty if nil
```
cam72cam: Above we os.Exit(1) if the blacklist can't be loaded.  Why add the nil check here?

Yantrio: Valid. removed.

src/internal/blacklist/blacklist.go
Yantrio: removed.

cam72cam (approved):

PR #2495, opened 2025-07-16
Better logging when missing checksum files

### LINKED ISSUE #280: Better error message for missing SHA256SUMS file

diofeher:
Resolves #280 

Right now, our code is showing errors on the Github issue by using a json file: 

- This is what is writing the file: https://github.com/opentofu/registry/blob/main/src/cmd/add-provider/main.go#L119
- The validation error is here: https://github.com/opentofu/registry/blob/main/src/cmd/add-provider/main.go#L115
- After that, we read from the `output.json` file, and comment on Github https://github.com/opentofu/registry/blob/main/.github/scripts/submit-provider.sh#L34

This makes difficult to return multiple errors to the Github interface since only one error is allowed. This would need to be refactored in order to support multiple error messages, because sometimes we allow missing Checksum files, depending if there's one valid release, so this would have no errors.

Since this is a small issue, I propose we don't spend too much time on it and just return a message that point to people that a checksum file can be missing in addition to the "No valid versions" message. At the same time, I'm logging specifically which releases are missing.

### DIFF
--- a/.github/scripts/submit-provider.sh
+++ b/.github/scripts/submit-provider.sh
@@ -61,6 +61,6 @@ git commit -s -m "Create provider ${namespace}/${name}"
 git push -u origin "${branch}"
 
 # Create pull request and update issue
-pr=$(gh pr create --title "${TITLE}" --body "Created ${jsonfile/../src/} for provider ${namespace}/${name}.  Closes #${NUMBER}.") #--assignee opentofu/core-engineers)
-gh issue comment "${NUMBER}" -b "Your submission has been validated and has moved on to the pull request phase (${pr}).  This issue has been locked."
+pr=$(gh pr create --title "${TITLE}" --body "Created ${jsonfile/../src/} for provider ${namespace}/${name}. Closes #${NUMBER}.") #--assignee opentofu/core-engineers)
+gh issue comment "${NUMBER}" -b "Your submission has been validated and has moved on to the pull request phase (${pr}). This issue has been locked."
 gh issue lock "${NUMBER}" -r resolved
--- a/.github/scripts/submit-provider.sh
+++ b/.github/scripts/submit-provider.sh
@@ -61,6 +61,6 @@ git commit -s -m "Create provider ${namespace}/${name}"
 git push -u origin "${branch}"
 
 # Create pull request and update issue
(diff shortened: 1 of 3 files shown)


yottta (approved):

src/internal/provider/version.go
```diff
@@ -46,8 +46,9 @@ func (p Provider) VersionFromTag(release string) (*Version, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	if checksums == nil {
-		logger.Warn("checksums not found in release, skipping...")
+		logger.Warn(fmt.Sprintf("checksums not found in release %s, skipping...", version))
```
cam72cam: The log line here already has version/release embedded in it.

cam72cam: See the Logger.With above

.github/scripts/submit-provider.sh
```diff
@@ -61,6 +61,6 @@ git commit -s -m "Create provider ${namespace}/${name}"
 git push -u origin "${branch}"
 
 # Create pull request and update issue
-pr=$(gh pr create --title "${TITLE}" --body "Created ${jsonfile/../src/} for provider ${namespace}/${name}.  Closes #${NUMBER}.") #--assignee opentofu/core-engineers)
-gh issue comment "${NUMBER}" -b "Your submission has been validated and has moved on to the pull request phase (${pr}).  This issue has been locked."
+pr=$(gh pr create --title "${TITLE}" --body "Created ${jsonfile/../src/} for provider ${namespace}/${name}. Closes #${NUMBER}.") #--assignee opentofu/core-engineers)
```
cam72cam: Why include these whitespace changes?

diofeher: Oh, I just added because it seemed to small to open a separated PR and it was related to this one (since this message is related when adding providers). And this whitespace bothers me when I look these opened PRs :D Do you want me to remove it?

src/internal/provider/version.go
diofeher: That's true, I forgot about that! Let me revert this one. Thanks

.github/scripts/submit-provider.sh
cam72cam: Yes please, it's unrelated to this work.

PR #2508, opened 2025-07-18
Add retries on 403 errors when fetching assets from github releases

Yantrio:
Lately, we've been seeing strange 403 errors when fetching assets from GitHub releases.

This PR adds a retry mechanism to handle these errors gracefully. 
We suspect that there is something going on with GitHub's rate limiting or access control that is causing these intermittent failures, and this is a bit of a last-ditch attempt until we can reproduce this locally 100% of the time.

The retry mechanism will attempt to fetch the asset up to three times before giving up, which should help mitigate the impact of these errors on our users.

### DIFF
--- a/src/internal/github/assets.go
+++ b/src/internal/github/assets.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
+	"time"
 )
 
 // DownloadAssetContents downloads the contents of the asset at the given URL and returns it directly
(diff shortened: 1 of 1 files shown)


diofeher (approved):

cam72cam:
In my testing, I have not seen this have an impact.  From what I can tell, we get pinned to a github cache server that is unhappy and it takes something like a poorly formatted auth token to unstick it?

cam72cam (changes requested):
I don't see this as helping given everything we know about this Github issue.

diofeher:
Closed as not needed since Github fixed their bug.

PR #2785, opened 2025-09-10
Use unique provider key branches to prevent conflicts

### LINKED ISSUE #2150: Submitting multiple keys for the same provider at the same time creates conflicting branches

The provider key submission job creates a branch named `provider-key-submission_$PROVIDER`.  With multiple keys submitted at the same time, these actions conflict.  We should add a suffix to the branch name to allow these concurrent submissions.

### OpenTofu Version

```shell
NA
```

NA

cam72cam:
Fixes #2150

### DIFF
--- a/.github/scripts/submit-provider-key.sh
+++ b/.github/scripts/submit-provider-key.sh
@@ -70,7 +70,7 @@ mkdir -p "$(dirname "${keyfile}")"
 mv tmp.key "${keyfile}"
 
 # Create Branch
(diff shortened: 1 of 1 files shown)


apparentlymart (approved):
I must admit that I'm not very familiar with the assumptions these scripts are making, but this seems plausible to me.

Presumably it's still possible to hit the error case if two submissions arrive very close together in time, but that seems very unlikely.

diofeher (approved):

PR #3651, opened 2026-02-21
ci: add concurrency groups to generate-and-deploy workflows

### LINKED ISSUE #1499: Add a concurrency group for generate and deploy

It looks like https://github.com/opentofu/registry/blob/main/.github/workflows/generate-and-deploy.yml could have multiple instances of that job syncing to R2 simultaneously.  I don't know if that will cause issues, but it should probably be avoided.

### OpenTofu Version

```shell
n/a
```

```hcl
// Snippet of HCL that can be used to help reproduce this problem
```

vikasrao23:
## Summary

Fixes #1499

Multiple simultaneous runs of the generate-and-deploy workflows can cause concurrent writes to the same R2 bucket, which may produce inconsistent state.

## Changes

Add concurrency groups to prevent overlapping runs:

- **`generate-and-deploy.yml`** and **`generate-and-deploy-with-delete.yml`**: share the `r2-sync` group with `cancel-in-progress: false` so a full sync is never cancelled mid-run, but a queued duplicate is dropped.
- **`generate-and-deploy-recent.yml`**: uses its own `r2-sync-recent` group with `cancel-in-progress: true` since it runs every 15 minutes and a stale queued run can safely be replaced by the next trigger.

## Testing

Workflow syntax validated locally. The concurrency keys ensure:
- Full syncs are never interrupted
- Duplicate queued runs are dropped
- Recent-only syncs can cancel stale runs

### DIFF
--- a/.github/workflows/generate-and-deploy-recent.yml
+++ b/.github/workflows/generate-and-deploy-recent.yml
@@ -13,6 +13,10 @@ on:
           - Development
           - Production
 
+concurrency:
+  group: r2-sync-recent
+  cancel-in-progress: true
+
 jobs:
   generate-and-sync:
     timeout-minutes: 10
--- a/.github/workflows/generate-and-deploy-with-delete.yml
+++ b/.github/workflows/generate-and-deploy-with-delete.yml
@@ -11,6 +11,10 @@ on:
           - Development
           - Production-with-delete
 
+concurrency:
+  group: r2-sync
+  cancel-in-progress: false
+
 jobs:
   generate-and-sync:
     runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


vikasrao23:
Thanks for the review! Working on addressing all the feedback now.

yottta (commented):
IMO, there should be just one concurrency key.

Since the concern here is around the possible negative side-effects of syncing information on R2 concurrently, all of the workflows should block on one another.
Even more, since the R2 access is the concern, this workflow that runs way more often than the others, could actually create the issues that we want to look out for so it would make even more sense to have it on the same key with no cancellation of the others.

PR #3744, opened 2026-03-05
Calculate and store provider meta (H1 hash and size)

cam72cam:
This is for https://github.com/opentofu/opentofu/pull/3434

Once merged, this will add the additional metadata for new provider releases the registry discovers. I am intentionally not adding the backfill process in this PR to keep the concerns separate.

This code has been running on https://github.com/cam72cam/opentofu-registry/ for a few months now and does not show any significant performance impediment.

Sample Generation:
v1/providers/hashicorp/local/0.1.0/download/linux/amd64
```json
{
  "protocols": [
    "5.0"
  ],
  "os": "linux",
  "arch": "amd64",
  "filename": "terraform-provider-local_0.1.0_linux_amd64.zip",
  "download_url": "https://github.com/opentofu/terraform-provider-local/releases/download/v0.1.0/terraform-provider-local_0.1.0_linux_amd64.zip",
  "shasums_url": "https://github.com/opentofu/terraform-provider-local/releases/download/v0.1.0/terraform-provider-local_0.1.0_SHA256SUMS",
  "shasums_signature_url": "https://github.com/opentofu/terraform-provider-local/releases/download/v0.1.0/terraform-provider-local_0.1.0_SHA256SUMS.sig",
  "shasum": "34d89a61574f2387e77c7d641823fce9e2a0339a1527d6a25404b2c3c7d10da7",
  "signing_keys": {
    "gpg_public_keys": [
      {
        "ascii_armor": "...",
        "key_id": "0C0AF313E5FD9F80"
      }
    ]
  },
  "packages": {
    "darwin_amd64": {
      "hashes": [
        "zh:5b30e49702c1e18200fda9628fcd36fe782c502125e25972153b3660ac1122f2",
        "h1:cPER98d6LWpUZ4QPNx1nH8a6vTyI6gQtojPhzBf1Rtw="
      ],
      "package_size": 3694015
    },
    "linux_386": {
      "hashes": [
        "zh:4b2cd9c6b21ce4142124a194ef96480aad4efdbcae4449d9678fc939c29d3754",
        "h1:/WX6cY2ijQM/le95WUqDeAq6JO5BRljkMEClcf1vGQY="
      ],
      "package_size": 3278244
    },
    "linux_amd64": {
      "hashes": [
        "zh:34d89a61574f2387e77c7d641823fce9e2a0339a1527d6a25404b2c3c7d10da7",
        "h1:MDZJm4NtP8OCv5e4SI3avLV8xZbQXVJOyLGNNmO7nS4="
      ],
      "package_size": 3516023
    },
    "windows_386": {
      "hashes": [
        "zh:a51ed8a7992aed9047547c4b6811c74b7dc7f2452ace313a790eeb929b01ade0",
        "h1:1ctrk5nQ+U0swj7as9vYeyKkMth6uDh7u5/wT1dkm1s="
      ],
      "package_size": 3266737
    },
    "windows_amd64": {
      "hashes": [
        "zh:80bc8857c4f069e392155214b6cdbe4c0e02924fe7cc621f30936cdd46ddc8b6",
        "h1:eiyrESXZ5jx0xnOt2/2aMqBfSZFa9QJNKO+75SnqaeA="
      ],
      "package_size": 3514114
    }
  }
}
```

<details>
<summary>Potential Alternative</summary>

v1/providers/hashicorp/local/versions
```json
{
  "versions": [
    {
      "version": "0.1.0",
      "protocols": [
        "5.0"
      ],
      "platforms": [
        {
          "os": "darwin",
          "arch": "amd64",
          "hashes": [
            "zh:5b30e49702c1e18200fda9628fcd36fe782c502125e25972153b3660ac1122f2",
            "h1:cPER98d6LWpUZ4QPNx1nH8a6vTyI6gQtojPhzBf1Rtw="
          ],
          "package_size": 3694015
        },
        {
          "os": "linux",
          "arch": "386",
          "hashes": [
            "zh:4b2cd9c6b21ce4142124a194ef96480aad4efdbcae4449d9678fc939c29d3754",
            "h1:/WX6cY2ijQM/le95WUqDeAq6JO5BRljkMEClcf1vGQY="
          ],
          "package_size": 3278244
        },
        {
          "os": "linux",
          "arch": "amd64",
          "hashes": [
            "zh:34d89a61574f2387e77c7d641823fce9e2a0339a1527d6a25404b2c3c7d10da7",
            "h1:MDZJm4NtP8OCv5e4SI3avLV8xZbQXVJOyLGNNmO7nS4="
          ],
          "package_size": 3516023
        },
        {
          "os": "windows",
          "arch": "386",
          "hashes": [
            "zh:a51ed8a7992aed9047547c4b6811c74b7dc7f2452ace313a790eeb929b01ade0",
            "h1:1ctrk5nQ+U0swj7as9vYeyKkMth6uDh7u5/wT1dkm1s="
          ],
          "package_size": 3266737
        },
        {
          "os": "windows",
          "arch": "amd64",
          "hashes": [
            "zh:80bc8857c4f069e392155214b6cdbe4c0e02924fe7cc621f30936cdd46ddc8b6",
            "h1:eiyrESXZ5jx0xnOt2/2aMqBfSZFa9QJNKO+75SnqaeA="
          ],
          "package_size": 3514114
        }
      ]
    }
  ]
}
```

</details>

### DIFF
--- a/src/internal/github/assets.go
+++ b/src/internal/github/assets.go
@@ -7,8 +7,14 @@ import (
 	"net/http"
 )
 
-// DownloadAssetContents downloads the contents of the asset at the given URL and returns it directly
+// DownloadAssetContents downloads the contents of the asset at the given URL and returns it directly (32kb limit)
 func (c Client) DownloadAssetContents(downloadURL string) ([]byte, error) {
+	return c.DownloadAssetContentsLimited(downloadURL, 32*1024) // 32kb is a reasonable limit for most metadata files
+}
+
+// DownloadAssetContentsLimited downloads the contents of the asset at the given URL and returns it directly, as long as it
+// is within the given limit.
+func (c Client) DownloadAssetContentsLimited(downloadURL string, limit int64) ([]byte, error) {
 	done := c.assetThrottle()
 	defer done()
 
@@ -30,9 +36,21 @@ func (c Client) DownloadAssetContents(downloadURL string) ([]byte, error) {
 		return nil, fmt.Errorf("unexpected status code when downloading asset %s: %d", downloadURL, resp.StatusCode)
 	}
 
-	contents, err := io.ReadAll(resp.Body)
+	if resp.ContentLength > limit {
+		return nil, fmt.Errorf("exceeded download file content length of %vb for %s with a reported content length of %vb", limit, downloadURL, resp.ContentLength)
+	}
+
+	limiter := &io.LimitedReader{
+		R: resp.Body,
+		N: limit,
+	}
+	contents, err := io.ReadAll(limiter)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read asset contents of %s: %w", downloadURL, err)
+	}
+
+	if limiter.N <= 0 {
+		return nil, fmt.Errorf("exceeded download file size of %vb for %s with a reported content length of %vb", limit, downloadURL, resp.ContentLength)
 	}
 
 	logger.Info("asset successfully downloaded", slog.Int("size", len(contents)))
--- a/src/internal/provider/version.go
+++ b/src/internal/provider/version.go
@@ -80,17 +80,24 @@ func (p Provider) VersionFromTag(release string) (*Version, error) {
 				DownloadURL: p.ArtifactURL(release, version, suffix),
 			}
 			target.SHASum, ok = checksums[target.Filename]
-			if ok {
-				v.Targets = append(v.Targets, target)
-				continue
+			if !ok {
+				// now try and pull it with the v in the version
+				target.Filename = p.ArtifactName("v"+version, suffix)
+				target.DownloadURL = p.ArtifactURL(release, "v"+version, suffix)
+				target.SHASum, ok = checksums[target.Filename]
+				if !ok {
+					// Release target without checksum (invalid)
+					continue
+				}
 			}
-			// now try and pull it with the v in the version
-			target.Filename = p.ArtifactName("v"+version, suffix)
-			target.DownloadURL = p.ArtifactURL(release, "v"+version, suffix)
-			target.SHASum, ok = checksums[target.Filename]
-			if ok {
-				v.Targets = append(v.Targets, target)
+
+			target.Hash1, err = p.CalculateHash1(target.DownloadURL, target.SHASum)
+			if err != nil {
+				// Release is inaccessable, misconfigured, or corrupt
+				return nil, err
 			}
+
+			v.Targets = append(v.Targets, target)
 		}
 	}
 
--- a/src/internal/provider/version.go
+++ b/src/internal/provider/version.go
@@ -91,9 +91,10 @@ func (p Provider) VersionFromTag(release string) (*Version, error) {
 				}
 			}
 
-			target.Hash1, err = p.CalculateHash1(target.DownloadURL, target.SHASum)
+			target.Hash1, target.Size, err = p.CalculateHash1AndSize(target.DownloadURL, target.SHASum)
 			if err != nil {
-				// Release is inaccessable, misconfigured, or corrupt
+				// Release is inaccessible, misconfigured, or corrupt
+				// TODO consider of this constitutes a failure, or if we should fall back to pre-h1 logic
 				return nil, err
 			}
 
--- a/src/internal/provider/version.go
+++ b/src/internal/provider/version.go
@@ -94,7 +94,7 @@ func (p Provider) VersionFromTag(release string) (*Version, error) {
 			target.Hash1, target.Size, err = p.CalculateHash1AndSize(target.DownloadURL, target.SHASum)
 			if err != nil {
 				// Release is inaccessible, misconfigured, or corrupt
-				// TODO consider of this constitutes a failure, or if we should fall back to pre-h1 logic
+				// TODO consider if this constitutes a failure, or if we should fall back to pre-h1 logic
 				return nil, err
 			}
 
--- a/src/internal/v1api/platform.go
+++ b/src/internal/v1api/platform.go
@@ -1,7 +1,1 @@
 package v1api
-
-// Platform represents a platform that a provider supports.
-type Platform struct {
-	OS   string `json:"os"`
-	Arch string `json:"arch"`
-}
--- a/src/internal/v1api/platform.go
+++ b/src/internal/v1api/platform.go
@@ -1,1 +1,0 @@
-package v1api
--- a/src/internal/provider/hashes.go
+++ b/src/internal/provider/hashes.go
@@ -1,0 +1,58 @@
+package provider
+
+import (
+	"archive/zip"
+	"bytes"
+	"crypto/sha256"
+	"fmt"
+	"io"
+
+	"golang.org/x/mod/sumdb/dirhash"
+)
+
+func (p Provider) CalculateHash1(releaseDownloadUrl string, shaExpected string) (string, error) {
+	contents, assetErr := p.Github.DownloadAssetContents(releaseDownloadUrl)
+	if assetErr != nil {
+		return "", fmt.Errorf("failed to download release for calcualting hashes: %w", assetErr)
+	}
+
+	shaHasher := sha256.New()
+	_, err := io.Copy(shaHasher, bytes.NewBuffer(contents))
+	if err != nil {
+		panic(err)
+	}
+
+	shaCalculated := fmt.Sprintf("%x", shaHasher.Sum(nil))
+	if shaCalculated != shaExpected {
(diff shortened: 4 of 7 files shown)


yottta (approved):
Thanks for sharing the repo that you had this running on. Way easier to reason about the changes when seen side by side with what it generates.

src/internal/v1api/platform.go
```diff
@@ -1,7 +1 @@
 package v1api
```
Yantrio: we can maybe remove this file now

src/internal/provider/version.go
```diff
@@ -80,17 +80,25 @@ func (p Provider) VersionFromTag(release string) (*Version, error) {
...
-			target.Filename = p.ArtifactName("v"+version, suffix)
-			target.DownloadURL = p.ArtifactURL(release, "v"+version, suffix)
-			target.SHASum, ok = checksums[target.Filename]
-			if ok {
-				v.Targets = append(v.Targets, target)
+
+			target.Hash1, target.Size, err = p.CalculateHash1AndSize(target.DownloadURL, target.SHASum)
+			if err != nil {
+				// Release is inaccessible, misconfigured, or corrupt
+				// TODO consider of this constitutes a failure, or if we should fall back to pre-h1 logic
```
Yantrio:
```suggestion
				// TODO consider if this constitutes a failure, or if we should fall back to pre-h1 logic
```

Yantrio (changes requested):
Overall I like the approach. 

I think now though that `DownloadAssetContents` should wrap it's reading in a io.LimitReader to ensure that memory is not a large problem now we're reading provider binaries.

What do you think?

apparentlymart:
Given that [our provider registry protocol](https://opentofu.org/docs/internals/provider-registry-protocol/) is an integration point we encourage others to implement, we should probably keep third-party implementations in mind as we finalize the protocol here.

Any situation where the same information is duplicated in multiple places is risky, because only the subset of the data that OpenTofu's provider installer interacts with is going to be covered when someone tests their third-party implementation, and if we then try to use other parts of the data in later releases we're likely to get interop problems with third parties that didn't correctly implement the parts that the client wasn't previously using.

My initial instinct is to put the big table of hashes in the objects returned from the "versions" endpoint, because we're intentionally describing _all_ of the available packages rather than just the one for the platform where the client is running. If we duplicate that same information into every variant of the "download" endpoint then it's possible that third-party implementations might make them mismatch in ways that cause the dependency lock file to get populated differently depending on what platform `tofu init` was run on, which would create a new variation of the problem we were trying to solve with all of this. :grinning: 

That does mean that the big list of versions will now be an even larger payload than it already was, but in the current protocol that's the only endpoint we have that describes the metadata OpenTofu should use _regardless of chosen platform_, so I think that's the best compromise and I expect gzip/etc compression will reduce the overhead significantly given that it will be quite repetitive.

apparentlymart:
Looking more closely at [the response format for "List Available Versions"](https://opentofu.org/docs/internals/provider-registry-protocol/#sample-response) I notice that we already have a data structure representing which platforms are supported, and so maybe we should just extend that to include the additional metadata:

```json
{
  "versions": [
    {
      "version": "2.0.0",
      "protocols": ["4.0", "5.1"],
      "platforms": [
        {
          "os": "darwin",
          "arch": "amd64",
          "hashes": [
            "h1:...",
            "zh:..."
          ],
          "size": 12345
        },
        {
          "os": "linux",
          "arch": "amd64",
          "hashes": [
            "h1:...",
            "zh:..."
          ],
          "size": 12345
        },
        {
          "os": "linux",
          "arch": "arm",
          "hashes": [
            "h1:...",
            "zh:..."
          ],
          "size": 12345
        },
        {
          "os": "windows",
          "arch": "amd64",
          "hashes": [
            "h1:...",
            "zh:..."
          ],
          "size": 12345
        }
      ]
    }
  ]
}
```

This does of course leave us with the platforms specified as separate `os` and `arch` values instead of having an object with properties like `"windows_amd64"`, but it seems like the most natural extension of the API we already had here, and avoids duplicating the information about which platforms are supported.

**Edit:** Immediately after I posted this I realized that the original comment included an example just like this for the "versions" endpoint. :man_facepalming: Sorry, I was still focused on the example of the "download" endpoint when I made the original comment, and forgot there was already an example for this one that matched what I was proposing. With that in mind, the essence of what I'm proposing here is to just leave the "download" endpoint unmodified and add this new information _only_ in the "versions" endpoint.

cam72cam:
Part of the problem is that OpenTofu's workflow does not directly connect this extended version data (through ListVersions) with query a given package (GetMeta).  We infact completely ignore the platforms field in the versions endpoint inside OpenTofu.

apparentlymart:
Yeah, the current structure of the client internally is annoying here because the "list versions" method literally just returns version numbers and nothing else.

I would propose that we change the internal design to match this API structure, rather than duplicating information in the documented API just to preserve an internal implementation detail. For example, we could change the internal representation of the "list versions" operation to return a set of objects representing versions and their metadata, and then have the "package meta" function take one whole version object from that response instead of just the version number.

Unfortunately [the network mirror protocol](https://opentofu.org/docs/internals/provider-network-mirror-protocol/) is already shaped a little differently in a way that better matches how the internal API is designed -- there's a "list versions" endpoint that only returns version numbers and then a single endpoint per version that describes all available packages for all platforms -- so I expect we'll need to also continue supporting the old shape where the initial response has no hashes and then `PackageMeta` decides the hashes. But I think better for our implementation details to be a little awkward if it makes the external protocol easier for third-parties to implement correctly.

src/internal/provider/version.go
cam72cam: [17338e4](https://github.com/opentofu/registry/pull/3744/commits/17338e4106d890eed0b082107d4c455f349bcfb4)

src/internal/v1api/platform.go
cam72cam: [17338e4](https://github.com/opentofu/registry/pull/3744/commits/17338e4106d890eed0b082107d4c455f349bcfb4)

cam72cam:
Based on a random sample, the generation is now 1.6x the current size.

The versions listing files are dramatically larger (~3x):
```
gen_main/v1/providers/hashicorp/azuread/versions 76K
gen_new/v1/providers/hashicorp/azuread/versions 264K
gen_main/v1/providers/hashicorp/helm/versions 48K
gen_new/v1/providers/hashicorp/helm/versions 160K
gen_main/v1/providers/hashicorp/tfcoremock/versions 12K
gen_new/v1/providers/hashicorp/tfcoremock/versions 32K
```

The individual version listing files are only 1.5x the original size, although there are more of them
```
4.0K    gen_main/v1/providers/hashicorp/azuread/2.50.0/download/linux/amd64
6.2K    gen_new/v1/providers/hashicorp/azuread/2.50.0/download/linux/amd64
```

From a client perspective, I think the original plan to include the platform data in the individual download responses is the more efficient one.

cam72cam:
@Yantrio I've implemented the file size limit in [0d09357](https://github.com/opentofu/registry/pull/3744/commits/0d093579b4db3167ea4dde6a6b903edf32c2f5e1).  Let me know what you think.

diofeher (approved):
Your work here looks great, @cam72cam!

My only issue with this PR is the lack of new tests added (mainly on hashes.go), but I guess we could add them in a subsequent PR.

src/internal/github/assets.go
```diff
@@ -30,11 +36,23 @@ func (c Client) DownloadAssetContents(downloadURL string) ([]byte, error) {
...
+	limiter := &io.LimitedReader{
+		R: resp.Body,
+		N: limit,
+	}
+	contents, err := io.ReadAll(limiter)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read asset contents of %s: %w", downloadURL, err)
 	}
 
+	if limiter.N <= 0 {
```
diofeher, line 52: I was reading LimitedReader's [documentation](https://pkg.go.dev/io#LimitedReader.Read), and I'm in doubt: 

```
A LimitedReader reads from R but limits the amount of data returned to just N bytes. Each call to Read updates N to reflect the new amount remaining. Read returns EOF when N <= 0 or when the underlying R returns EOF.
```

Wouldn't this code be redundant since we're already checking the error from `io.ReadAll` above?

Yantrio: I dont think this is redundant. the LimitedReader will send an `EOF` when it gets to the limit.

This is the same as reaching the end of the file, and so `io.Reader` doesn't see this as an error, but a success case.

SO, with that in mind, i think we need to check `limiter.N` to see if we reached the limit (by checking the remaining limit being greater than 0)

cam72cam: That is correct. EOF a special error that has a particular meaning within the IO world and is handled correctly here.

PR #3751, opened 2026-03-05
Backfill provider meta

cam72cam:
Depends on https://github.com/opentofu/registry/pull/3744

Introduces a soft-timeout in the backfill process, it won't start any new version backfills once the deadline is exceeded.  At most a version backfill takes 30s (usually 1-2s) and won't exceed our expected action runtime.

### DIFF
--- a/src/cmd/bump-versions/main.go
+++ b/src/cmd/bump-versions/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"log/slog"
 	"os"
+	"time"
 
 	"github.com/opentofu/registry-stable/internal/blacklist"
 	"github.com/opentofu/registry-stable/internal/github"
@@ -13,6 +14,8 @@ import (
 )
 
 func main() {
+	started := time.Now()
+
 	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
 	logger.Info("Starting version bump process for modules and providers")
 
@@ -20,6 +23,7 @@ func main() {
 	moduleNamespace := flag.String("module-namespace", "", "Which module namespace to limit the command to")
 	providerDataDir := flag.String("provider-data", "../providers", "Directory containing the provider data")
 	providerNamespace := flag.String("provider-namespace", "", "Which provider namespace to limit the command to")
+	targetDuration := flag.Duration("target-duration", time.Minute*5, "Used to limit how much of a backfill this command can perform")
 
 	flag.Parse()
 
@@ -67,4 +71,26 @@ func main() {
 	}
 
 	logger.Info("Completed version bump process for modules and providers")
+
+	deadline := started.Add(*targetDuration)
+	remainingTime := deadline.Sub(time.Now())
+
+	ctx, _ = context.WithTimeout(ctx, remainingTime)
+
+	if ctx.Err() != nil {
+		logger.Info("Skipping backfill process, deadline exceeded")
+		return
+	}
+
+	logger.Info("Beginning backfill process for providers", slog.Any("time_allocated", remainingTime))
+
+	err = providers.Parallel(20, func(p provider.Provider) error {
+		if ctx.Err() != nil {
+			// Outta-time
+			return nil
+		}
+		return p.BackfillVersionData(ctx)
+	})
+
+	logger.Info("Completed backfill process for providers")
 }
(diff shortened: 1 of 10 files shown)


src/internal/provider/backfill.go
```diff
@@ -0,0 +1,51 @@
...
+				break
+			}
+		}
+		if hasMeta {
+			continue
+		}
+
+		newVersion, err := p.VersionFromTag("v" + version.Version)
+		if err != nil {
+			// TODO log
```
diofeher: Can you implement that log? I think it would be better together with this PR.

src/cmd/bump-versions/main.go
```diff
@@ -67,4 +71,26 @@ func main() {
...
+	if ctx.Err() != nil {
+		logger.Info("Skipping backfill process, deadline exceeded")
+		return
+	}
+
+	logger.Info("Beginning backfill process for providers", slog.Any("time_allocated", remainingTime))
+
+	err = providers.Parallel(20, func(p provider.Provider) error {
+		if ctx.Err() != nil {
+			// Outta-time
```
diofeher, line 109: Can you add a better context for this timeout error skipping here? 

Maybe with a log or a comment "we're good to skip because we don't need to download all providers..."

src/internal/provider/build.go
```diff
@@ -119,6 +120,12 @@ func (p Provider) buildMetadata() (*Metadata, error) {
 		}
 		// append the new release to the metadata
 		meta.Versions = append(meta.Versions, *result.v)
+
+		addedReleases = true
+	}
+	if !addedReleases {
+		// Prevent file modification if not changed
+		return nil, nil
```
diofeher, line 128: What do you think about an error raised here instead?

I feel this `nil, nil` can get confused with other returns above... But I'm not super strong about this one

src/cmd/bump-versions/main.go
```diff
@@ -67,4 +71,26 @@ func main() {
 	}
 
 	logger.Info("Completed version bump process for modules and providers")
+
+	deadline := started.Add(*targetDuration)
+	remainingTime := deadline.Sub(time.Now())
+
+	ctx, _ = context.WithTimeout(ctx, remainingTime)
```
diofeher: We need to cancel the timeout here. Now that I'm thinking, a static linter should've caught that...

diofeher (changes requested):

src/cmd/bump-versions/main.go
cam72cam: Given the current structure of this PR, we are simply using the context as a timing signal.

for completion, I could add the typical defer here in case we start using the context as a "real" context.

src/cmd/bump-versions/main.go
cam72cam: We log a message two lines up saying we are beginning the backfill with a given allocated time. If we exceed that time, we should not continue.

src/internal/provider/build.go
cam72cam: This is a strong convention in this codebase already.  There is no result and there is no error.

src/cmd/bump-versions/main.go
diofeher: Oh, I was mostly commenting on this regarding the possible go leak when you don't call `defer cancel`.

yottta: If it fails earlier, yeah, that will leak the go routine behind. But obviously, not a big issue since the agent it runs on is kinda ephemeral.
But for good practices, `defer cancel()` shouldn't hurt.

src/cmd/bump-versions/main.go
cam72cam: Addressed in [43b3a06](https://github.com/opentofu/registry/pull/3751/commits/43b3a06fe1928a24adba648de35c4a4aae17cf7d)

src/cmd/bump-versions/main.go
cam72cam: Addressed in [43b3a06](https://github.com/opentofu/registry/pull/3751/commits/43b3a06fe1928a24adba648de35c4a4aae17cf7d)

src/internal/provider/backfill.go
cam72cam: [43b3a06](https://github.com/opentofu/registry/pull/3751/commits/43b3a06fe1928a24adba648de35c4a4aae17cf7d)

diofeher (approved):
Thanks for the fixes! There are some lint errors, but I'll approve so you can merge it later ^^

yottta (approved):
Looks clean! Thanks for this

PR #3759, opened 2026-03-05
Add lint on PRs

diofeher:
Based on that [comment](https://github.com/opentofu/registry/pull/3751#discussion_r2892281843), I noticed that we do not have linters for this repo on PRs.


src/internal/github/client.go
```diff
@@ -76,9 +76,8 @@ func (c Client) WithLogger(log *slog.Logger) Client {
 // transport is a http.RoundTripper that makes sure all requests have the
 // correct User-Agent and Authorization headers set.
 type transport struct {
-	token  string
-	ctx    context.Context
-	parent http.Transport
+	token string
+	ctx   context.Context
```
apparentlymart, line 80: I guess this was because `parent` is unused?

It's interesting that `transport.RoundTrip` seems to instantiate a new `http.Transport` to use as "parent" on each call, but it doesn't seem like there's any variation in how it gets instantiated and so it probably _could_ be stored in a field and reused.

Of course that's out of scope for this PR since we should just focus on lint changes here and not try to make any substantial changes, but I just thought it was interesting and wonder if there was an interesting story behind the current way it's structured. 😀

apparentlymart (approved):
With the set of linters reduced down to a more reasonable set, this seems like it found some good things to fix!

It's nice that in this newer codebase we can actually follow our policy of using the default golangci-lint configuration, without having to make weird exceptions for legacy code. 😀

.github/workflows/verify.yml
```diff
@@ -0,0 +1,25 @@
+name: Verify
+permissions: {}
+on:
+  pull_request:
+
+jobs:
```
yottta, line 8: Without no rules here, wouldn't this run also on community submissions?

diofeher: Good point! Let me change it to be more specific when `.go` code is changed, thanks for the catch!

diofeher:
I missed the DCO on the last commit, but I added it when squashing and merging to main. It fails on the PR but it's ok on main.

PR #3762, opened 2026-03-06
Upgrade action lib versions

diofeher:
- Upgrade checkout **action version** to v6
- Upgrade setup-go **action version** to v6
- Upgrade golangci-lint **action version** to v9 and golanci-lint **version** to v2.10

Fixed lints and some typos on the way.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -5,6 +5,9 @@ on:
     paths:
       - 'src/**'
 
+env:
+  GOLANGCI_LINT_VERSION: v2.10
+
 jobs:
   lint:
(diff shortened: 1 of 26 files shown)


diofeher:
As discussed with the team, we decided to go with the `golangci-lint` defaults.

yottta (commented):
Looks good in general and the bumps, even if upgrades over a lot of versions, should be fine considering the nature of the updated actions.
One question though, did you check that the changed error messages does not disrupt something else? Error checks, comments posted on submissions on failures/success?

yottta (approved):

PR #3806, opened 2026-03-12
feat(terraform-provider-utils): Add cloudposse/utils provider versions 1.32.0 and 1.33.0

johncblandii:
## Summary

Add missing versions 1.32.0 and 1.33.0 of the `cloudposse/utils` provider to the registry dataset.

These versions were published as GitHub releases but were not picked up by the automated sync, likely because they are new releases on the 1.x branch published after the 2.x line already exists. The [[releases RSS feed](https://github.com/cloudposse/terraform-provider-utils/releases.atom)](https://github.com/cloudposse/terraform-provider-utils/releases.atom) includes both versions, but the sync appears to have skipped them.

## Versions added

- **v1.32.0** — [[GitHub Release](https://github.com/cloudposse/terraform-provider-utils/releases/tag/v1.32.0)](https://github.com/cloudposse/terraform-provider-utils/releases/tag/v1.32.0) (published 2026-03-11)
- **v1.33.0** — [[GitHub Release](https://github.com/cloudposse/terraform-provider-utils/releases/tag/v1.33.0)](https://github.com/cloudposse/terraform-provider-utils/releases/tag/v1.33.0) (published 2026-03-12)

Both releases have signed SHA256SUMS and binary artifacts for 11 platform targets (darwin, freebsd, linux, windows across amd64/arm64/386/arm).

## Why the sync may have missed these

The provider maintains parallel 1.x and 2.x release lines. The registry already has v2.1.0 (released the same day as v1.32.0), suggesting the sync logic may not expect new releases on an older major version after a newer major version exists. Additionally, the releases RSS feed contains floating tags (`v1`, `v2`) alongside proper semver tags, which could interfere with version parsing.

## Changes

- Updated `providers/c/cloudposse/utils.json` to include v1.32.0 and v1.33.0 entries with shasums, download URLs, and file sizes sourced from the GitHub releases.
- `h1` hashes are left empty, consistent with the majority of existing entries in this file.

PR #3807, opened 2026-03-12
Track provider bump errors

cam72cam:
This resolves issues like those seen in https://github.com/opentofu/registry/pull/3806 where "older" version numbers are not detected.  We originally added the shouldUpdateMetadataFile check to ignore old erroring versions, but it was a workaround that has run it's course.

It uses an exponential backoff approach:
```
delay = 2^NumErrorsRecorded * 15 minutes
nextAttemptSoonest = latestAtteptTime + delay
```
* Delay1 = 2^1 * N = 2N
* Delay2 = 2^2 * N = 4N
* Delay3 = 3^2 * N = 8N
* DelayZ = Z^2 * N = 2^ZN
where N = 15 minutes

The only major issue is that this approach will update a bunch of json meta files in the repo over the first few bumps and cause lots of sync traffic.  Given what we saw with the recent backfill, I'm not terribly worried about that.

### DIFF
--- a/src/internal/provider/build.go
+++ b/src/internal/provider/build.go
@@ -1,9 +1,12 @@
 package provider
 
 import (
+	"errors"
 	"fmt"
 	"log/slog"
+	"math"
 	"slices"
+	"time"
 
 	"github.com/opentofu/registry-stable/internal"
 	"github.com/opentofu/registry-stable/internal/blacklist"
@@ -21,6 +24,7 @@ func (meta Metadata) filterNewReleases(releases []string, namespace, name string
 
 	var newReleases = make([]string, 0)
 	var blacklistedCount = 0
+	var erroredCount = 0
 	for _, r := range releases {
 		version := internal.TrimTagPrefix(r)
 		if !existingVersions[version] {
@@ -34,12 +38,42 @@ func (meta Metadata) filterNewReleases(releases []string, namespace, name string
 				blacklistedCount++
 				continue
 			}
-			// only append the release if it does not already exist in the metadata and is not blacklisted
+
+			var errorEntries []VersionError
+			var latestError time.Time
+			for _, errored := range meta.VersionErrors {
+				if errored.Version == version {
+					if errored.UTCTime.After(latestError) {
+						latestError = errored.UTCTime
+					}
+					errorEntries = append(errorEntries, errored)
+				}
+			}
+
+			isErrored := false
+			if len(errorEntries) > 0 {
+				// Simple doubling backoff
+				dur := time.Minute * 15 * time.Duration(math.Pow(float64(len(errorEntries)), 2))
+				if latestError.Add(dur).After(time.Now()) {
+					isErrored = true
+				}
+			}
+
+			if isErrored {
+				meta.Logger.Warn("Skipping errored version",
+					slog.String("namespace", namespace),
+					slog.String("name", name),
+					slog.String("version", version))
+				erroredCount++
+				continue
+			}
+
+			// only append the release if it does not already exist in the metadata, is not errored, and is not blacklisted
 			newReleases = append(newReleases, r)
 		}
 	}
 
-	meta.Logger.Info(fmt.Sprintf("Found %d releases that do not already exist in the metadata file (%d blacklisted)", len(newReleases), blacklistedCount))
+	meta.Logger.Info(fmt.Sprintf("Found %d releases that do not already exist in the metadata file (%d blacklisted, %d errored)", len(newReleases), blacklistedCount, erroredCount))
 
 	return newReleases
 }
@@ -93,6 +127,7 @@ func (p Provider) buildMetadata() (*Metadata, error) {
 	}
 
 	type versionResult struct {
+		r   string
 		v   *Version
 		err error
 	}
@@ -104,15 +139,26 @@ func (p Provider) buildMetadata() (*Metadata, error) {
 		r := r
 		go func() {
 			version, err := p.VersionFromTag(r)
-			verChan <- versionResult{version, err}
+			verChan <- versionResult{r, version, err}
 		}()
 	}
 
 	addedReleases := false
+	var errs []error
 	for range newReleases {
 		result := <-verChan
 		if result.err != nil {
-			return nil, result.err
+			var nonFatal ErrVersionNonFatal
+			if errors.As(result.err, &nonFatal) {
+				meta.VersionErrors = append(meta.VersionErrors, VersionError{
+					Version: internal.TrimTagPrefix(result.r),
+					Message: nonFatal.Error(),
+					UTCTime: time.Now().UTC(),
+				})
+				continue
+			}
+			errs = append(errs, result.err)
+			continue
 		}
 		if result.v == nil {
 			// Not a valid release, skipping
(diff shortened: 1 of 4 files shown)


johncblandii:
I think #3808 was intended to update the provider ref but it has been a while now and it still doesn't show v1.32.0: https://search.opentofu.org/provider/cloudposse/utils/latest.

src/internal/provider/build.go
```diff
@@ -34,12 +38,42 @@ func (meta Metadata) filterNewReleases(releases []string, namespace, name string
...
+					}
+					errorEntries = append(errorEntries, errored)
+				}
+			}
+
+			isErrored := false
+			if len(errorEntries) > 0 {
+				// Simple doubling backoff
+				dur := time.Minute * 15 * time.Duration(math.Pow(float64(len(errorEntries)), 2))
+				if latestError.Add(dur).After(time.Now()) {
```
yottta, line 57:
```suggestion
				if latestError.Add(dur).After(time.Now().UTC()) {
```
Just to be sure

src/internal/provider/version.go
```diff
@@ -65,7 +65,7 @@ func (p Provider) VersionFromTag(release string) (*Version, error) {
 
 	if checksums == nil {
 		logger.Warn("checksums not found in release, skipping...")
-		return nil, nil
+		return nil, ErrVersionNonFatal{fmt.Errorf("checksums not found in release")}
```
yottta, line 68: I am not sure that I get this. Do we record this error on missing checksums for a particular release.
Why would we want to retry this? A release without a checksum is pretty much invalid. Do we want to double check in case the author updated[^1] the release with the checksums later?

[^1]: I am not sure that we should allow and _encourage_ such an approach. We should try as much as possible to treat every release as immutable and ignore the ones that are just invalid

src/internal/provider/build.go
```diff
@@ -34,12 +38,42 @@ func (meta Metadata) filterNewReleases(releases []string, namespace, name string
 				blacklistedCount++
 				continue
 			}
-			// only append the release if it does not already exist in the metadata and is not blacklisted
+
+			var errorEntries []VersionError
```
yottta: This could be just an int. Not needed to hold the whole slice.

src/internal/provider/build.go
```diff
@@ -104,15 +131,26 @@ func (p Provider) buildMetadata() (*Metadata, error) {
...
 		if result.err != nil {
-			return nil, result.err
+			var nonFatal ErrVersionNonFatal
+			if errors.As(result.err, &nonFatal) {
+				meta.VersionErrors = append(meta.VersionErrors, VersionError{
+					Version: internal.TrimTagPrefix(result.r),
+					Message: nonFatal.Error(),
+					UTCTime: time.Now().UTC(),
+				})
+				continue
```
yottta, line 151:
```suggestion
                addedReleases = true
				continue
```
In cases where **all** the versions failed, the information will be lost because without `addedReleases`  being true, it returns nil meta which skips the writing of the file.

Maybe we could also rename the `addedReleases` to more specific (`metadataUpdated`)

src/internal/provider/version.go
```diff
@@ -126,3 +126,15 @@ func (p Provider) VersionFromTag(release string) (*Version, error) {
 
 	return &v, nil
 }
+
+type ErrVersionNonFatal struct {
```
yottta, line 130:
```suggestion
type ErrRetryableVersion struct {
```

yottta: Or we could just have a sentinel error and use that for wrapping and just call `errors.Is`. 🤷 
It's more of a matter of taste in this case.

src/internal/provider/update.go
```diff
@@ -1,90 +1,19 @@
...
 
-	p.Logger.Info("Could not find latest tag, updating...", slog.String("tag", semVerTag))
-	return true, nil
-}
-
-// getSemVerTagsFromRSS returns a list of semver tags from the RSS feed
-// ignoring all non-valid semver tags
-func (p Provider) getSemVerTagsFromRSS() ([]string, error) {
-	releasesRssUrl := p.RSSURL()
-	tags, err := p.Github.GetTagsFromRSS(releasesRssUrl)
```
yottta, line 59: Maybe we can cleanup `Github.GetTagsFromRSS` too, but we can also leave it since it's part of the client and maybe we might need it later.

Just wanted to highlight it. Up to you.

yottta:
> I think #3808 was intended to update the provider ref but it has been a while now and it still doesn't show v1.32.0: https://search.opentofu.org/provider/cloudposse/utils/latest.

In the meantime, it appeared: https://search.opentofu.org/provider/cloudposse/utils/v1.32.0

src/internal/provider/version.go
cam72cam: The main reason we re-try is that a user might publish a release, and then fix it once they realize that it is improperly configured.

This path allows us to record that error (potentially to show in the registry UI), and check later to see if the release has been fixed.

src/internal/provider/build.go
cam72cam: Internally to the time package, this does not matter.  Time.After compares this correctly, double checked in the impl.

src/internal/provider/version.go
cam72cam: I decided not to use a sentinel error as we may expand what constitutes a non-fatal error.

src/internal/provider/update.go
cam72cam: Our approach thus far has been to leave most of the unused functionality in the Github package alone in case we need it again.  Most of the code that was removed here is just version sorting.

I'd like to do a more thorough cleanup of this codebase at some point, but I'm very hesitant as this is battle tested code.

src/internal/provider/build.go
cam72cam: Good catch, fixed in 628538e389

src/internal/provider/build.go
cam72cam: Fixed in 628538e389

src/internal/provider/version.go
yottta: Yeah, but still not sure that we should promote the idea of updating releases 🤷 
But you've been more on this than I did so I will leave this to you.

yottta (approved):

Yantrio (approved):

src/internal/provider/build.go
```diff
@@ -121,17 +160,30 @@ func (p Provider) buildMetadata() (*Metadata, error) {
 		// append the new release to the metadata
 		meta.Versions = append(meta.Versions, *result.v)
 
-		addedReleases = true
+		// remove any error version records now that we have had a successful result
+		meta.VersionErrors = slices.DeleteFunc(meta.VersionErrors, func(errored VersionError) bool {
+			return errored.Version == result.r
```
diofeher: There's a possible bug here:

`errored.Version` comes from the struct which contains:

```
Version: internal.TrimTagPrefix(result.r)
```

So it contains the version without the `v` prefix.

While the `result.r` comes from:

```
			version, err := p.VersionFromTag(r)
			verChan <- versionResult{r, version, err}
```

`VersionFromTag` calls `TrimTagPrefix` at https://github.com/opentofu/registry/blob/0b0b573db5474b3587a63cd01d5c247ceee67119/src/internal/provider/version.go#L40

The right way to check wouldn't be:

`errored.Version` == `result.version`?

cam72cam: Good catch @diofeher.  I've addressed this in aaeafa36ef

cam72cam: This is my biggest regret with this codebase.  The version and tag types should be distinct aliases of string to prevent this sort of mistake.

diofeher (approved):

PR #3931, opened 2026-03-26
 deps: sha-pin all github actions

diofeher:
SHA-pinning all actions used by this repo. I decided not to update them to maintain the same functionality, but only to match the existing releases.


apparentlymart (approved):
I did not re-verify that the commits you chose here match the tags we were previously using, but switching to specific commits seems like a good idea!

yottta (approved):
Checked the commit hashes against the tag ones and all looks good. Thanks

Issue #3974, opened 2026-03-31
Notice: GoReleaser v2.15.0+ breaks provider checksum signature verification


Yantrio:
GoReleaser v2.15.0+ has a bug that breaks checksum signature verification, directly impacting OpenTofu/Terraform provider authors.

### What's happening

Starting in v2.15.0, `.sig` files (including `checksums.txt.sig`) are added back into `checksums.txt` after signing, invalidating the signature. The fix attempt in v2.15.1 did not seem to fully resolve it.

- **v2.15.0** — broken with `signs.artifacts: all` ([goreleaser/goreleaser#6508](https://github.com/goreleaser/goreleaser/issues/6508))
- **v2.15.1** — still broken with `signs.artifacts: checksum` ([goreleaser/goreleaser#6514](https://github.com/goreleaser/goreleaser/issues/6514))
- **v2.14.3** — last known working version

### Recommended action

Pin GoReleaser to **v2.14.3** until the upstream fix lands.

If you've already released a provider version using GoReleaser v2.15.0+, **do not re-create or overwrite that release**. Instead, pin GoReleaser to v2.14.3 and ship a **new patch version** of your provider with valid checksums.

The OpenTofu Registry handles this gracefully, existing indexed versions will continue to work, and we will **not** be re-indexing releases for this issue.

Upstream maintainers are [actively working on a fix](https://github.com/goreleaser/goreleaser/issues/6514#issuecomment-4163136687).

flbla:
the latest version of goreleaser : 2.15.2 fixed it
https://github.com/goreleaser/goreleaser/releases/tag/v2.15.2

Yantrio:
I'll close this now that this is fixed.

PR #3987, opened 2026-04-02
Remove version impacted by go-releaser bug for re-index

### LINKED ISSUE #3985: Request re-indexing: volcengine/volcenginecc versions 0.0.35, 0.0.36, 0.0.37
### Problem
We are the maintainers of `volcengine/terraform-provider-volcenginecc`. Due to the GoReleaser checksum bug (related to #3974), our releases for versions **0.0.35**, **0.0.36**, and **0.0.37** were initially published with invalid checksums.
The Terraform Registry rejected these releases, but the OpenTofu Registry indexed them successfully with the broken checksums. We then fixed our GoReleaser configuration (pinned to a working version) and **re-created the GitHub releases** for these three versions with valid checksums and signatures.
As a result, the OpenTofu Registry now has **stale/invalid checksum data** for these versions that does not match the current GitHub releases. Users attempting to install these versions via OpenTofu will encounter checksum verification failures.
### Affected versions
| Provider | Namespace | Version |
|----------|-----------|---------|
| volcenginecc | volcengine | 0.0.35 |
| volcenginecc | volcengine | 0.0.36 |
| volcenginecc | volcengine | 0.0.37 |
### Request
Could the OpenTofu team please **re-index** these three versions so the registry reflects the corrected checksums from the current GitHub releases? 
Alternatively, if re-indexing is not possible, could these versions be **removed from the registry data** so they can be re-indexed on the next sync cycle?
### Provider Repository
https://github.com/volcengine/terraform-provider-volcenginecc
### Additional Context
- We are public members of the `volcengine` organization
- The GPG signing key is already registered with the OpenTofu Registry
- All three re-released versions have valid checksums and signatures

cam72cam:
Related to: #3974
Fixes:  #3985

It should go out in the next bump/generate cycle.


Yantrio:
Closing as per my comment here https://github.com/opentofu/registry/issues/3985#issuecomment-4177478114

PR #4093, opened 2026-04-18
ci: auto-run workflows on PR through pull_request_target

diofeher:
`module` and `provider` submissions are not automatically triggering the workflows. This is an attempt to solve that.

### DIFF
--- a/.github/scripts/verify-pr-json.sh
+++ b/.github/scripts/verify-pr-json.sh
@@ -10,6 +10,9 @@ fi
 failed=0
 error_file="/tmp/verify-errors.md"
 
+# Allowed chars in path components fed to shell.
+safe_path_re='^[A-Za-z0-9_.-]+$'
+
 write_error_header() {
 	if [[ ! -f "${error_file}" ]]; then
 		cat >"${error_file}" <<'HEADER'
@@ -42,6 +45,13 @@ for file in ${changed_providers}; do
 	# Path format: providers/{first_char}/{namespace}/{name}.json
 	namespace=$(echo "${file}" | cut -d'/' -f3)
 	name=$(basename "${file}" .json)
+
+	if ! [[ "${namespace}" =~ ${safe_path_re} ]] || ! [[ "${name}" =~ ${safe_path_re} ]]; then
+		echo "::error::Invalid provider path: ${file}"
+		failed=1
+		continue
+	fi
+
 	repo="${namespace}/terraform-provider-${name}"
 
 	tmpdir=$(mktemp -d)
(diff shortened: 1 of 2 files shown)


.github/scripts/verify-pr-json.sh
```diff
@@ -10,6 +10,9 @@ fi
 failed=0
 error_file="/tmp/verify-errors.md"
 
+# Allowed chars in path components fed to shell.
+safe_path_re='^[A-Za-z0-9_.-]+$'
```
apparentlymart, line 14: Does this pattern enforce the same rules as the registry itself expects for what is a valid namespace and name?

OpenTofu itself allows more characters than what you've included here to appear in the namespace and name parts of source addresses, but the design intention for that is only to make sure it's possible for a registry to _choose_ to allow non-ASCII characters in names... it's up to each registry to decide its own policy for what subset of names to allow within what OpenTofu accepts. So I'm not meaning to say here that this is definitely wrong, but just checking that we're not accidentally making the naming restrictions stricter than they were before.

.github/workflows/verify-pr-json.yaml
```diff
@@ -1,10 +1,16 @@
 name: Verify PR JSON
 
 on:
-  pull_request:
+  pull_request_target:
```
bendiknesbo, line 4: I have not reviewed this PR, but I'm sharing that using `pull_request_target` opens up for potential attacks, if misconfigured: https://orca.security/resources/blog/pull-request-nightmare-github-actions-rce/

diofeher: Thanks for the feedback @bendiknesbo 

We're aware of that, that's why we're restricting the usage of it

bendiknesbo: Great! Then feel free to dismiss/resolve this comment :)

.github/scripts/verify-pr-json.sh
diofeher: That's a good point, but this regex is actually loosen then the one being accepted by the registry repo:

https://github.com/opentofu/registry/blob/97f3862bb704cb0cd57aa8bfbe5cfcb6875e9c1d/src/cmd/add-provider/main.go#L58-L59

https://github.com/opentofu/registry/blob/97f3862bb704cb0cd57aa8bfbe5cfcb6875e9c1d/src/cmd/add-module/main.go#L58-L59

So I'm just adding another layer of defense...

apparentlymart: Great, thanks for checking that!

diofeher:
Converting to draft as I would like to test another approach first.

Issue #4103, opened 2026-04-20
JSON PR check should ignore error's `utc_time`


yottta:
### Thank you for reporting an issue.

As seen [here](https://github.com/opentofu/registry/pull/4102), the `Verify PR JSON` fails because of the `utc_time` field in the recently added `versions_errors`.

It's obvious that by running the repository parsing in 2 separate time instances, will generate different time of the reported error.

We should ignore that field when comparing the generated JSON during the PR check with the one submitted in the PR file.

### OpenTofu Version

```shell
N/A
```

### OpenTofu Configuration Files

N/A

PR #4113, opened 2026-04-21
Record module and provider discover time

cam72cam:
Also serve module downloads by commit hash

* https://github.com/opentofu/opentofu/issues/3990
* https://github.com/opentofu/opentofu/issues/1545
* https://github.com/opentofu/opentofu/issues/1942

I decided to use today's date as the "we first discovered this module/provider".  By the time we have this supported in a opentofu release, we should have plenty of historical data (months).

One potential alternative is to leave the date field in the metadata blank and choose a sane default (like above) in the v1api generator.

Another alternative is to simply omit dates that are unknown and let opentofu / operators decide how they wish to handle that.  This option would mean that historical releases would never be available for the advanced latest version (by date) selection described in 3990

As currently implemented, this will cause *large* backfill commits and may slow down the sync changes job for the first hour after merge.

### DIFF
--- a/src/internal/github/git.go
+++ b/src/internal/github/git.go
@@ -8,8 +8,13 @@ import (
 	"strings"
 )
 
-func parseTagsFromStdout(lines []string) ([]string, error) {
-	tags := make([]string, 0, len(lines))
+type Tag struct {
+	Ref    string
+	Commit string
+}
+
+func parseTagsFromStdout(lines []string) ([]Tag, error) {
+	tags := make([]Tag, 0, len(lines))
 
 	for _, line := range lines {
 		if !strings.Contains(line, "refs/tags/") {
@@ -31,14 +36,14 @@ func parseTagsFromStdout(lines []string) ([]string, error) {
 			return nil, fmt.Errorf("invalid format for tag '%s', no version provided", line)
 		}
 
-		tags = append(tags, tag)
+		tags = append(tags, Tag{Ref: tag, Commit: fields[0]})
 	}
 
 	return tags, nil
 }
 
 // GetTags lists the tags of the remote repository and returns the refs/tags/ found
-func (c Client) GetTags(repositoryURL string) ([]string, error) {
+func (c Client) GetTags(repositoryURL string) ([]Tag, error) {
 	done := c.cliThrottle()
 	defer done()
 
--- a/src/internal/github/git.go
+++ b/src/internal/github/git.go
@@ -2,6 +2,7 @@ package github
 
 import (
 	"bytes"
+	"encoding/hex"
 	"fmt"
 	"log/slog"
 	"os/exec"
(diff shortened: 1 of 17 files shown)


apparentlymart:
Since our registry is just one implementation of the registry protocol, our client-side code is going to need to have _some_ rule for dealing with registries that don't have this metadata and so we could potentially decide to treat existing items in our registry the same way and just leave them without a timestamp at all, treating them in whatever way we decide to treat entries from existing third-party registries that don't return timestamps.

However, if we're defining the reference timestamp as whenever the registry learned about something for the first time then we're effectively asserting that something was published _no later than_ the given timestamp, and so backfilling whatever is the current time when we run the backfill seems consistent with that: we do know that these things were definitely published before that time, or else they wouldn't be in our registry dataset already.

So with all of that said: I expect we'll document this new API property in the protocol docs as a conservative "published no later than" date rather than actually a publication date, and so backfilling it seems like the best treatment.

src/internal/github/git.go
```diff
@@ -31,14 +36,14 @@ func parseTagsFromStdout(lines []string) ([]string, error) {
 			return nil, fmt.Errorf("invalid format for tag '%s', no version provided", line)
 		}
 
-		tags = append(tags, tag)
+		tags = append(tags, Tag{Ref: tag, Commit: fields[0]})
```
yottta: We have plenty of checks in this function regarding the correctness of the parsed information, including "if the tag is empty coming from the `strings.Fields` results".
Just to keep consistency, I would add the same check for the `fields[0]` too.

src/internal/module/build.go
```diff
@@ -63,7 +65,11 @@ func (m Module) BuildMetadata() (*Metadata, error) {
 				blacklistedCount++
 				continue
 			}
-			meta.Versions = append(meta.Versions, Version{Version: t})
+			meta.Versions = append(meta.Versions, Version{
+				Version:    t.Ref,
+				Commit:     t.Commit,
+				Discovered: new(time.Now().UTC()),
```
yottta, line 71: Since we store here `time.Now()`, please update the [`verify-pr-json.sh`](https://github.com/opentofu/registry/blob/main/.github/scripts/verify-pr-json.sh#L29) script diff to ignore this field too.
Otherwise, we will end up in a similar situation with what was reported in #4103.

src/internal/module/backfill.go
```diff
@@ -0,0 +1,76 @@
...
+	var errs []error
+	madeChanges := false
+	var backfilled, skipped, errored int
+	for key, version := range meta.Versions {
+		if err := ctx.Err(); err != nil {
+			errs = append(errs, err)
+			break
+		}
+
+		if version.Discovered != nil && version.Commit != "" {
```
yottta, line 32: I know that this is for the backfill, but this check here makes me wonder about the following, in the grand scheme of things:

What if the release has been re-issued? We know instances where maintainers of modules/providers decided to recreate a particular version.
I know that we don't encourage that, but that doesn't mean that this is gonna change the maintainers' behavior about this.

src/internal/module/module.go
```diff
@@ -51,7 +65,12 @@ func (m Module) RSSURL() string {
 // the file should just contain a link to GitHub to download the tarball, ie:
 // git::https://github.com/terraform-aws-modules/terraform-aws-iam?ref=v5.30.0
 func (m Module) VersionDownloadURL(version Version) string {
-	return fmt.Sprintf("git::%s?ref=%s", m.RepositoryURL(), version.Version)
+	ref := version.Commit
+	if ref == "" {
+		// TODO remove this fallback once backfill is complete
```
yottta, line 70: And update the comment
```suggestion
		// TODO remove this fallback once backfill is complete and update the method comment.
```

lbordowitz:
Question: do we just change the `BackfillVersionData` method wholesale depending on what operation we want to do that time? I'm asking because the `provider.BackfillVersionData` method is kinda being "torn out" and replaced with code to find `Discovered` times for relevant commits.

Editing to kinda answer my own question: this `backfill.go` file dates to [the recent past](https://github.com/opentofu/registry/pull/3751/changes), so it's not like this is a hard-and-fast "thing we do" type thing.

cam72cam:
Yep, we should have cleared out that function after the last backfill was completed.

src/internal/module/backfill.go
cam72cam: IMO we would remove that release from the json file and PR that change.  That would re-index it with the current date as the release date.  It sucks, but it does accomplish the idea that module releases are immutable.

yottta (approved):

src/internal/module/backfill.go
```diff
@@ -0,0 +1,76 @@
...
+					break
+				}
+			}
+			if !foundRelease {
+				err := fmt.Errorf("release not found for version %s", version.Version)
+				m.Logger.Error("Failed to backfill version", slog.String("version", version.Version), slog.Any("err", err))
+				errs = append(errs, err)
+				errored++
+			}
+		}
```
Yantrio, line 60: should we break/continue here?

src/internal/provider/version.go
```diff
@@ -31,20 +33,21 @@ func (p Provider) ArtifactName(version string, suffix string) string {
...
+	version := internal.TrimTagPrefix(release.Ref)
 
-	logger := p.Logger.With(slog.String("release", release))
+	logger := p.Logger.With(slog.String("release", release.Ref))
 
 	v := Version{
 		Version:             version,
 		SHASumsURL:          p.ArtifactURL(release, version, "SHA256SUMS"),
 		SHASumsSignatureURL: p.ArtifactURL(release, version, "SHA256SUMS.sig"),
+		Discovered:          new(time.Now()),
```
Yantrio: https://github.com/opentofu/registry/pull/4113/changes#r3122120032

This comment made me realise this, we need .UTC() here.
```suggestion
		Discovered:          new(time.Now().UTC()),
```

Yantrio (changes requested):
2 small nits, but otherwise seems good to me 👍

src/internal/module/backfill.go
cam72cam: 7a87aa54f3

src/internal/provider/version.go
cam72cam: 7a87aa54f3

Yantrio (approved):

PR #4237, opened 2026-05-14
Add: warn submitter when no GPG key found on provider submission

### LINKED ISSUE #115: If somebody submits a provider and we do not currently have the GPG key, prompt them to also add that

It is a 2 step process to upload gpg keys right now, we should prompt people to also upload their public keys where possible.

JESUSLUG:
Adds a warning comment to provider submission issues when no GPG key is found for the submitted namespace or provider, informing the submitter to follow the 2-step process.

Closes #115

### DIFF
--- a/src/cmd/check-provider-gpg/main.go
+++ b/src/cmd/check-provider-gpg/main.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log/slog"
+	"os"
+
+	regaddr "github.com/opentofu/registry-address"
+	"github.com/opentofu/registry-stable/internal/files"
+	"github.com/opentofu/registry-stable/internal/gpg"
+)
+
+type Output struct {
+	HasKeys bool   `json:"has_keys"`
+	Message string `json:"message"`
+}
+
+func main() {
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
+
+	namespace := flag.String("namespace", "", "The provider namespace to check")
+	name := flag.String("name", "", "The provider name to check")
+	outputFile := flag.String("output", "", "Path to write JSON result to")
+	gpgDataDir := flag.String("gpg-data", "", "Directory containing the GPG key data")
+
+	flag.Parse()
+
+	if *namespace == "" || *name == "" || *gpgDataDir == "" {
+		logger.Error("--namespace, --name, and --gpg-data are required")
+		os.Exit(1)
+	}
+
+	collection := gpg.KeyCollection{
+		Namespace:    *namespace,
+		ProviderName: *name,
+		Directory:    *gpgDataDir,
+	}
+
+	keys, err := collection.ListKeys()
+	if err != nil {
+		logger.Error("Failed to list GPG keys", slog.Any("err", err))
+		os.Exit(1)
+	}
+
+	output := Output{HasKeys: len(keys) > 0}
+
+	if !output.HasKeys {
+		provider, providerErr := regaddr.ParseProviderSource(fmt.Sprintf("%s/%s", *namespace, *name))
+		if providerErr != nil {
+			logger.Error("Failed to parse provider source", slog.Any("err", providerErr))
+			os.Exit(1)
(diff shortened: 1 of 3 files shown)


src/cmd/check-provider-gpg/main.go
```diff
@@ -0,0 +1,67 @@
...
+	keys, err := collection.ListKeys()
+	if err != nil {
+		logger.Error("Failed to list GPG keys", slog.Any("err", err))
+		os.Exit(1)
+	}
+
+	output := Output{HasKeys: len(keys) > 0}
+
+	if !output.HasKeys {
+		fqn := fmt.Sprintf("registry.opentofu.org/%s/%s", *namespace, *name)
```
yottta: Please use [`regaddr.ParseProviderSource` ](https://github.com/opentofu/registry-address/blob/45f3562374e4c9a94057c70d9ea57fcfb2b86bb1/provider.go#L219)instead. This is where the default registry domain comes from.
You can see an example [here](https://github.com/opentofu/registry/blob/cded2be8cbd93b78cd8d632a38aea7757031d0e5/src/cmd/add-provider/main.go#L73).

yottta (commented):
Thanks for the test addition on the gpg package.
Just one comment added regarding with how the FQN is generated..

src/cmd/check-provider-gpg/main.go
JESUSLUG: Hii. fixed using regaddr.ParseProviderSource now.
Thanks for the feedback

yottta (changes requested):
Please remove  the `check-provider-gpg` binary.

JESUSLUG:
> Please remove the `check-provider-gpg` binary.

Sorry, the binary slipped in when I compiled locally.  This is ready

yottta (approved):
Thanks. That looks good to me.
Let's wait for other maintainers to check this out.

src/cmd/check-provider-gpg/main.go
```diff
@@ -0,0 +1,72 @@
...
+	output := Output{HasKeys: len(keys) > 0}
+
+	if !output.HasKeys {
+		provider, providerErr := regaddr.ParseProviderSource(fmt.Sprintf("%s/%s", *namespace, *name))
+		if providerErr != nil {
+			logger.Error("Failed to parse provider source", slog.Any("err", providerErr))
+			os.Exit(1)
+		}
+		output.Message = fmt.Sprintf(
+			"No GPG key found for the \"%s\" provider and neither for the \"%s\" organisation. "+
```
Yantrio: One tiny tiny suggestion/nitpick:
```suggestion
			"No GPG key found for the \"%s\" provider nor for the \"%s\" namespace. "+
```

src/cmd/check-provider-gpg/main.go
```diff
@@ -0,0 +1,72 @@
...
+			provider.String(), *namespace,
+		)
+	}
+
+	if *outputFile != "" {
+		if jsonErr := files.SafeWriteObjectToJSONFile(*outputFile, output); jsonErr != nil {
+			logger.Error("Failed to write output file", slog.Any("err", jsonErr))
+			os.Exit(1)
+		}
+	}
```
Yantrio: What if we moved this inside the `output.HasKeys` check above? Then we only write this file if the key does not exist?

JESUSLUG:
Hi. Friendly ping on this PR 👀

yottta:
> Hi. Friendly ping on this PR 👀

The other maintainers will take a look at this next week. Thanks for your patience.

Yantrio (changes requested):
LGTM oveall, I just have a couple tiny nitpicks.

JESUSLUG:
Hi @Yantrio ! I've addressed the changes mentioned above. 
Thank you for the feedback!

Yantrio (approved):

yottta:
Screenshot 2026-05-25 at 08 41 54

Works like a charm. Thanks for the help @JESUSLUG! 🙌

PR #4294, opened 2026-05-22
Fix shell script as it fails during linting

yottta:
Screenshot 2026-05-22 at 20 21 53


diofeher (approved):

apparentlymart (approved):
I think this lint is telling us something valid -- that if the `jq` command fails for any reason then we'd end up asking `gh` to post an empty comment on the issue. But an empty comment would presumably be rejected with an error anyway, so checking ourselves beforehand would just cause it to return a different errror.

Issue #4337, opened 2026-05-28
Unclear lag between providers releasing a new version and it appearing in the registry


joe-a-dbt:
### Thank you for reporting an issue.

I am trying to figure out why https://github.com/fivetran/terraform-provider-fivetran/releases/tag/v1.9.31 (which was released over an hour ago at the time I'm writing this) did not yet appear in the OpenTofu registry, which is still showing v1.9.30 as the latest. https://github.com/fivetran/terraform-provider-fivetran/releases/tag/v1.9.31 was released at 10:47AM CDT but https://github.com/opentofu/registry/commit/540486436877ca05d28e2346b4d1146f07bb945a did not run until 12:03PM CDT (76 minutes later). Additionally, https://search.opentofu.org/provider/fivetran/fivetran/latest is still showing 1.9.30 as the latest as of 12:27PM CDT (100 minutes later). This appears to be a significant variance from https://github.com/opentofu/registry/blob/main/PROCEDURES.md#user-complaint-provider-version-is-not-available for maintainer troubleshooting, which sets a 30 min expectation.

Additionally, beyond this immediate instance I'm flagging, I also want to flag the need for improving the documentation in the README. It would be great if the README or a more use-facing document than the PROCEDURES set a clear expectation for how long it should take between a provider releasing a new version and the OpenTofu registry being updated.

### OpenTofu Version

```shell
N/A
```

### OpenTofu Configuration Files

N/A

joe-a-dbt:
Looks like 1.9.31 showed up in the last few minutes. But I still think some investigation is warranted about why it took so long and clarifying the documentation.

Yantrio:
I understand the frustration here of the delay, it's something that we're discussing internally. I'd love to explain a little bit more about what we're seeing here though.

## Intended flow
The registry works on a three-tiered system of GitHub Actions. These are in charge of three data-sets.

- Registry metadata in this repository
- API Data in cloudflare r2
- search.opentofu.org documentation scraping

All 3 of these tiers act independantly through GitHub Actions across both github.com/opentofu/registry and github.com/opentofu/registry-ui

**The first tier** aims to run every 15 minutes , and will check github repositories for new releases and commit that to this repository. This is what you are seeing generating [this commit](https://github.com/opentofu/registry/commit/540486436877ca05d28e2346b4d1146f07bb945a). 
Action: https://github.com/opentofu/registry/blob/main/.github/workflows/bump-versions.yml#L11
[...]

PR #4379, opened 2026-06-04
Bump versions now triggers generate-and-deploy

Yantrio:
We've been noticing lately that github scheduling is getting flaky. This is the first step in an attempt to mitigate some of these issues.

After this if its improving, we will attempt the following

- Also trigger the `index` job in github.com/opentofu/registry-ui
- Move cron scheduling to an external mechanism (For now, cloudflare workers)

Then, this means we no longer rely on github action scheduling.

This is based on the investigation I did here: https://github.com/opentofu/registry/issues/4337#issuecomment-4567128219

---

Note, this also removes the concept of the development env for the registry as this has not been used in a LONG time.

### DIFF
--- a/.github/workflows/bump-versions.yml
+++ b/.github/workflows/bump-versions.yml
@@ -1,8 +1,6 @@
 name: Bump Provider and Module Versions
 
 # Use global lock
-#
-# https://github.com/opentofu/registry/pull/272#discussion_r1511031778
 concurrency:
   group: main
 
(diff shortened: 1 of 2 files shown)


.github/workflows/bump-versions.yml
```diff
@@ -1,30 +1,21 @@
 name: Bump Provider and Module Versions
 
 # Use global lock
-#
-# https://github.com/opentofu/registry/pull/272#discussion_r1511031778
```
yottta, line 5: I wouldn't remove it.

yottta (approved):

PR #634, opened 2024-06-04
Fixes #633: removing Linode provider versions 2.21.1 and 2.21.2

ghost:
Triggering a resync to fix #633. **⚠ This should be merged and then a rsync triggered immediately after. ⚠** 2.21.2 is included to make sure we are not triggering any possible ordering bugs.

PR #4811, opened 2026-07-29
Allow provider authors to request re-indexing a version

### LINKED ISSUE #642: Allow provider authors to request re-indexing a provider version

Currently, when a provider re-releases a binary, the old checksum stays in the registry. While normally provider versions should be immutable, this is often not the case. (See #630, #633, databricks/terraform-provider-databricks#3660, etc.) Since the Terraform registry apparently doesn't treat versions as immutable, we should have a way for provider authors to request re-indexing a specific version.

I propose that we create a GitHub workflow that lets a provider author request reindexing of a version if a) they are an org member in the organization and b) they have submitted a GPG key with the registry and c) the provider version verifies correctly against the GPG key. (This should be part of the issue verification process.) This should be sufficient to protect against most supply chain attacks since the binary will be verified against their GPG key. The process would involve manually merging the change that removes the given version from the registry data so we still have a level of manual control. The created PR should look roughly like #634.

aliamerj:
Closes #642.

Lets a provider author request a re-index of a specific version — e.g. if they
accidentally re-released a binary under an existing tag and the registry's
stored checksums are now stale.

## How it works

1. Author opens an issue using the new "Request Provider Version Re-index" template.
2. The workflow runs `verify-reindex-request`, which checks:
   - they're a public member of the provider's GitHub org
   - a GPG key is already on file for the provider
   - the version's *current* release assets still verify against that key
3. If all three pass, `remove-provider-version` deletes the stale entry and
   opens this PR automatically. The next `bump-versions` run re-fetches and
   re-verifies the version from scratch.

**This still requires a human to merge the PR and trigger a resync
afterward**, same as #634 — nothing here auto-merges

### DIFF
--- a/.github/ISSUE_TEMPLATE/reindex.yml
+++ b/.github/ISSUE_TEMPLATE/reindex.yml
@@ -1,0 +1,48 @@
+# IMPORTANT: Submissions MUST be made through the GitHub issue form UI, not via
+# the API, gh CLI, or by manually creating issues. The automated validation and
+# processing pipeline depends on the structured data from the issue form.
+name: Request Provider Version Re-index
+description: Ask the registry to re-verify and re-index a specific provider version, e.g. after a release was rebuilt under the same tag
+title: "Re-index: "
+labels: ["reindex", "submission"]
+body:
+  - type: markdown
+    attributes:
+      value: >-
+        This re-downloads and re-verifies the version's current release assets against your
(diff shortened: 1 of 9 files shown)


Yantrio:
Thanks for the submission, but we have decided to go in another direction with https://github.com/opentofu/registry/pull/4879.

PR #4853, opened 2026-08-02
fix: Prevent git ls-remote authentication prompts

### LINKED ISSUE #48: Handle git ls-remote asking for authentication
If a repository is removed or made private, git ls-remote will hang asking for user authentication and break the bumper.

mvanhorn:
Update `Client.GetTags` in `src/internal/github/git.go` so the spawned `git ls-remote` process receives `GIT_TERMINAL_PROMPT=0`, while retaining the rest of the parent process environment needed to locate Git and use configured non-interactive credential helpers. Keep the existing command arguments, throttling, stdout parsing, and stderr-rich error wrapping unchanged; once prompting is disabled, inaccessible repositories should follow the existing error path instead of blocking. Add subprocess-focused coverage in `src/internal/github/git_test.go` using a controlled fake `git` executable to observe the environment passed by `GetTags`, avoiding network access and avoiding a production abstraction created only for testing.

`Client.GetTags` runs `git ls-remote` directly to discover module and provider tags. When a repository has been removed or made private, Git may prompt for credentials on an interactive terminal, causing a local bumper run to wait indefinitely instead of returning an error. The issue is accepted, and maintainers later identified a solution in `opentofu/libregistry` that should be ported back to this repository. The existing module and provider callers already handle `GetTags` errors by skipping the affected repository, so the missing behavior belongs at the Git subprocess boundary.

With a controlled `git` executable that emits a valid tag, `GetTags` passes `GIT_TERMINAL_PROMPT=0` and still parses and returns the tag; When the parent test environment sets `GIT_TERMINAL_PROMPT` to an enabling value, the child process still observes `0`, proving local caller configuration cannot re-enable an interactive prompt.

Fixes #48

### DIFF
--- a/src/internal/github/git.go
+++ b/src/internal/github/git.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"log/slog"
+	"os"
 	"os/exec"
 	"strings"
 )
@@ -59,6 +60,7 @@ func (c Client) GetTags(repositoryURL string) ([]Tag, error) {
 	var buf bytes.Buffer
 	var bufErr bytes.Buffer
 	cmd := exec.Command("git", "ls-remote", "--tags", "--refs", repositoryURL)
+	cmd.Env = append(os.Environ(), "GIT_TERMINAL_PROMPT=0")
 	cmd.Stdout = &buf
 	cmd.Stderr = &bufErr
 
--- a/src/internal/github/git_test.go
+++ b/src/internal/github/git_test.go
@@ -1,10 +1,38 @@
 package github
 
 import (
+	"io"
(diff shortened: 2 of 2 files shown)


Yantrio:
Thanks for the submission @mvanhorn but I am going to close this for a couple reasons.

1. I suspect that this goes against our current AI policy in this repository: https://github.com/opentofu/registry/blob/main/AI-USAGE-POLICY.md

2. Upon looking at the issue, I actually think that this issue is no longer affecting anyone as nobody is running the registry locally. 

In the future, I recommend you reach out to us and ask to be assigned to an issue before any code is written to save future occurrences of this.

PR #4879, opened 2026-08-06
Introduce new policy around version removal

Yantrio:
It is becoming increasingly more common for provider and module authors to be requesting removals or re-indexing of specific versions. Internally the maintainers have often discussed the concept of versions being immutable to reduce any security risks.

This commit will bring in a new policy in which the maintainers state clearly that we will not be re-indexing or removing provider or module versions unless it is for extreme circumstance.

We are also introducing a new form for version removal that hopefully state clearly the policy, and direct people towards the POLICY.md section for further reading.

---

Note for reviewers, Im asking for human reviews of policies here, I've used an LLM to neaten up my wording in the POLICY.md, and to generate the issue templates. This PR requires 2 human reviews to be merged as per our AI policy in this repo.

Assisted-by: GitHub Copilot (Model: Claude Sonnet 5)


.github/ISSUE_TEMPLATE/version-reindex-request.yml
```diff
@@ -0,0 +1,77 @@
...
+      label: Repository
+      description: The full repository path, e.g. opentofu/terraform-provider-aws or GoogleCloudPlatform/terraform-google-secured-data-warehouse
+      placeholder: namespace/repository-name
+    validations:
+      required: true
+  - type: input
+    id: version
+    attributes:
+      label: Affected Version(s)
+      description: The exact version(s) you want re-indexed, e.g. 1.2.3
```
yottta:
```suggestion
      description: The exact version(s) you want re-indexed, e.g. 1.2.3, 1.2.4
```

.github/ISSUE_TEMPLATE/version-reindex-request.yml
```diff
@@ -0,0 +1,77 @@
...
+      required: false
+  - type: checkboxes
+    id: acknowledgement
+    attributes:
+      label: Acknowledgement
+      options:
+        - label: I have read the [Version Immutability Policy](https://github.com/opentofu/registry/blob/main/POLICY.md#version-immutability) and understand this request will only be actioned in extreme, case-by-case circumstances.
+          required: true
+        - label: This is not a request to re-index a version solely because of a CI mistake, duplicate publish, or a minor bug that could instead be fixed by releasing a new version.
+          required: true
```
yottta:
```suggestion
      options:
        - label: I have read the [Version Immutability Policy](https://github.com/opentofu/registry/blob/main/POLICY.md#version-immutability) and understand this request will only be actioned in extreme, case-by-case circumstances.
          required: true
        - label: This is not a request to re-index a version solely because of a CI mistake, duplicate publish, or a minor bug that could instead be fixed by releasing a new version.
          required: true
        - label: The GitHub organisation for which I open this request is [publicly visible in my GitHub profile](https://docs.github.com/en/account-and-profile/how-tos/organization-membership/publicizing-or-hiding-organization-membership) which will allow OpenTofu maintainers confirm the validity of this request.
          required: true
```

.github/ISSUE_TEMPLATE/version-reindex-request.yml
```diff
@@ -0,0 +1,77 @@
...
+labels: ["version-request"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        The OpenTofu Registry treats published versions as immutable. We do **not** re-index a version because of a CI mistake, a double-publish, or a minor bug. The correct fix for those is to release a new version in your own repository; the registry will pick it up automatically.
+
+        Before filling this out, please read the [Version Immutability Policy](https://github.com/opentofu/registry/blob/main/POLICY.md#version-immutability). Re-indexing is only considered on a case-by-case basis for extreme circumstances (e.g. a compromised signing key that makes verification of the original artifact impossible, or a release causing extreme disruption). Requests that don't meet this bar will be closed and redirected to release a new version instead.
+
+        **Make your case as strongly and specifically as possible.** Vague or unsupported requests are unlikely to be actioned.
```
yottta:
```suggestion
        **Make your case as strongly and specifically as possible.** Vague or unsupported requests are unlikely to be actioned.
        
        This request can be made only by a maintainer of the provider/module. The proof for that consists in having public visibility of the requester membership in the GitHub organisation that hosts the provider/module for which this request is created.
```

yottta: Same for `version-removal-request.yml`.

.github/ISSUE_TEMPLATE/version-reindex-request.yml
yottta: Same for `version-removal-request.yml`.

POLICY.md
```diff
@@ -16,7 +16,33 @@ The following categories of modules and providers will not be included in the Op
...
+
+Once a provider or module version has been published and indexed, the OpenTofu Registry treats it as immutable. Consumers and automation rely on the assumption that a given version number always resolves to the same artifacts and checksums; re-indexing or removing a version on demand undermines that trust for everyone who depends on it, not just the requester.
+
+For this reason, the OpenTofu Registry will **not** re-index or remove a version for reasons such as:
+
+- A CI pipeline built and published the release more than once.
+- A minor bug, regression, or mistake was discovered after release.
+- A provider or module author simply changed their mind about a release.
+
+In all of these cases, the correct fix is for the provider or module author to release a new version in their own repository. Consumers who need the fix can then upgrade to it, while anyone already relying on the original version is not disrupted by it changing or disappearing out from under them.
```
yottta:
```suggestion
In all of these cases, the correct fix is for the provider or module author to release a new version in their own repository. Consumers who need the fix can then upgrade to it, while anyone already relying on the original version is not disrupted by it changing or disappearing.
```

Maybe? The initial version sounds a little bit weird. 
But hey, you're the native speaker here 😅 So feel free to ignore this.

PROCEDURES.md
```diff
@@ -50,6 +50,17 @@ You can debug this on behalf of the user by performing the following steps:
...
 4. Run `gpg --verify ...SHA256SUMS.sig ...SHA256SUMS`
 
+## Version removal or re-indexing requests
+
+The [Version Immutability policy](POLICY.md#version-immutability) explains that we do not remove or re-index versions for routine reasons (CI double-builds, minor bugs, changed minds). Requests come in through the [Version Removal Request](https://github.com/opentofu/registry/issues/new?template=version-removal-request.yml) and [Version Re-indexing Request](https://github.com/opentofu/registry/issues/new?template=version-reindex-request.yml) issue templates.
+
+When triaging one of these issues:
+
+1. Check whether the stated reason matches one of the routine cases the policy explicitly rules out. If so, close the issue and point the author to the policy, recommending they release a new version in their own repository instead.
+2. If the request cites something in the exceptional category (malicious code, leaked secrets, compromised signing key, legal requirement, extreme disruption), verify the evidence provided (advisory links, CVEs, etc.) before acting.
```
yottta:
```suggestion
2. If the request cites something in the exceptional category, verify the evidence provided before acting.
```
I don't think that we should list all of those again since there is also a link to the policy already in this chapter.
Listing those just adds more maintenance burden and divergence risk between these 2 markdown files.

yottta (changes requested):
Thank you very much for this! 🥇 🚀

.github/ISSUE_TEMPLATE/version-reindex-request.yml
yottta: Same for `version-removal-request.yml`.

Yantrio: I was 50/50 on this one. I think recently the hashicorp/aws provider de-listed a version that was breaking a lot of things and it was the community who reached out to us to remove this. Maybe we dont make that one required?

yottta: And allow non-members of the organisation to request this? 🤔 
In the end, even if it's one of the exceptional reasons for which the removal/reindexation is requested, I still think that organisation members should drive this process. It's not OpenTofu's responsibility to analyse and flag provider/module versions that might be affected by the exceptional cases.
And if this request would be raised by a non-member, then what? Should OpenTofu maintainers get in touch with the maintainers of the provider/module for further investigation and confirmation?

I don't want to say that I deny your take on this, but I just want to understand what would be the course of action in cases like this and maybe include that somewhere in this PR.

apparentlymart: In the case of `hashicorp/aws`, for the sake of our registry the OpenTofu project itself is the effective maintainer and so we can interpret what we did for that recent broken release as the OpenTofu maintainers (who are also effectively the provider maintainers) deciding to pull a release based on feedback from users of that provider.

I would expect in that case for a provider user experiencing a problem to open an issue about it in the `opentofu/terraform-provider-aws` repository and then the provider maintainers (currently us) to decide whether to request removal.

I don't think anything else needs to be special about the treatment of these providers except that currently the maintainers of the providers are also maintainers of the registry. The process for removal need not be different beyond that, right?

POLICY.md
```diff
@@ -16,7 +16,33 @@ The following categories of modules and providers will not be included in the Op
...
+
+The registry may remove or re-index a version on a case-by-case basis for exceptional circumstances, including but not limited to:
```
apparentlymart: When we discussed this the other day I thought we had consensus that we would only allow removal and never allow modifying an already-published release in place.

Did I misunderstand the consensus or did something else change in the meantime to cause us to introduce the option of modifying a release without republishing it?

(We briefly talked about how in exceptional cases we might allow someone to publish a version with build metadata like `1.0.0+security.1` and then unpublish the original `1.0.0` if publishing a new patch release is impossible/impractical, but modifying an existing release _without_ modifying the version number like this would undermine the dependency lock file mechanism and undermine assumptions made by any provider mirrors that are acting as read-through caches over the upstream registry.)

cam72cam: That was my understanding as well, that we would not allow re-indexing moving forward.

Yantrio: I misunderstood and I will correct this. Thanks for pointing that out. Im happy to go with this approach

.github/ISSUE_TEMPLATE/version-reindex-request.yml
Yantrio: Im happy with that approach. In that case I'll take @yottta 's feedback onboard here. 

Thanks for talking it through!

yottta (approved):

cam72cam (approved):

Yantrio:
I will hold on merging this until @apparentlymart approves too as he had input before.

POLICY.md
```diff
@@ -16,7 +16,33 @@ The following categories of modules and providers will not be included in the Op
...
+- A minor bug, regression, or mistake was discovered after release.
+- A provider or module author simply changed their mind about a release.
+
+In all of these cases, the correct fix is for the provider or module author to release a new version in their own repository. Consumers who need the fix can then upgrade to it, while anyone already relying on the original version is not disrupted by it changing or disappearing.
+
+The registry may remove a version on a case-by-case basis for exceptional circumstances, including but not limited to:
+
+- The release contains malicious code, malware, or evidence of a supply chain compromise.
+- The release discloses secrets, credentials, or other sensitive information that must be revoked or scrubbed.
+- The signing key used for the release has been compromised, making the originally indexed checksums impossible to verify going forward.
```
apparentlymart, line 37: In this case, would we consider revoking the key that was compromised, rather than revoking the provider versions that were signed with it?

To be clear, I expect the result would be similar: those providers would still fail to install, just for the reason of an invalid key rather than the release being unpublished. But revoking the key seems overall simpler than trying to figure out all of the individual versions that a particular key was signed with an retracting them individually. 🤔

apparentlymart (approved):

POLICY.md
Yantrio: I think so, do you think we should explicitly state it in this policy or just use common-sense as this happens?

apparentlymart: Well, part of the reason I asked this question is that I wasn't sure if we actually _can_ yank a previously-published provider key in this way.

I'm glad to hear that it's possible, and indeed I don't think we need to mention this in `POLICY.md` in particular, but if this is something we expect to be doing then I'd like to document in `PROCEDURES.md` exactly how we would do that, since I don't personally feel confident in doing it without accidentally breaking something.

If the documentation for that would be time-consuming to prepare though, I don't mind leaving it to a separate PR since the main point of this one is to publish the provider-developer-facing policy, and only secondarily to explain to ourselves how we implement that policy. :grinning:

Issue #5150, opened 2026-08-24
Module: microsoftexpert/terraform-azurerm-system-center-virtual-machine-manager-virtual-machine-instance-guest-agent


microsoftexpert:
### Module Repository

microsoftexpert/terraform-azurerm-system-center-virtual-machine-manager-virtual-machine-instance-guest-agent

yottta:
Closing as this was covered by #5361.

PR #5361, opened 2026-09-01
Bulk import stubs for all modules in "microsoftexpert" GitHub account

### LINKED ISSUE #5323: Request to provide a process to bulk upload 1810 terraform modules to OpenTofu
Per @Yantrio request to me, I am opening a new issue requesting that @opentofu/maintainers see if they can come up with a process to allow me to bulk upload 1,810 terraform modules to OpenTofu.  Currently rules require us to submit modules one at a time but if there was a way that @opentofu/maintainers could allow me access even temporarily, or provide me an alternative process/interface/way to bulk upload all my modules, that is what I am asking for.

I am just trying to give back tot he community with great terraform modules.
My github site which all my modules are public is https://github.com/microsoftexpert

apparentlymart:
Instead of manually approving each of these modules individually, we'll add all of these stubs (based on GitHub's listing of all relevant repositories in the "microsoftexpert" user account) and then let the sync process populate their versions automatically after.

Closes https://github.com/opentofu/registry/issues/5323


apparentlymart:
One consequence of the one-off script I used here is that it bypasses the validation rule that made our automation reject https://github.com/opentofu/registry/issues/5150.

I'm not sure if having a module name longer than the submission script permits will cause problems for the background sync script. Perhaps I should pull that one out of here so that we can discuss the consequences of that long module name separately in its own issue. :thinking:

yottta (approved):
I would say to get this merged and let users/author of the module report any issue it might find with any of these.

